PR #98, opened 2026-04-13
#57: ELR-trampoline infrastructure for Pi 5 secondary-CPU preemption

### LINKED ISSUE #57: Secondary CPU timer preemption (Pi 5 / Jetson)
## Description

Enable timer-driven preemptive scheduling on secondary CPUs (currently only CPU 0 supports timer preemption).

From `docs/future-work.md` Section 7 and `docs/pi5-secondary-cpu-preemption.md`:

### Problem
On Pi 5 and Jetson, secondary CPU idle tasks use WFE (cooperative scheduling) instead of timer-driven WFI because timer ISR calling `schedule()` inside the exception handler abandons the exception frame (ELR_EL1/SPSR_EL1 for the interrupted task is lost).

### Proposed Solution: ELR_EL1 Trampoline
1. Timer ISR sets a "reschedule pending" flag instead of calling `schedule()` directly
2. Before ERET, check the flag
3. If set, modify ELR_EL1 to point to a trampoline that calls `schedule()` in thread context
4. Trampoline saves callee-saved registers, calls `schedule()`, then returns to original PC

### Scope
- Implement ELR_EL1 trampoline in assembly (`kernel/arch/arm64/vectors.S`)
- Enable timer on secondary CPUs (`idle_task_func` changes)
- Update test suite: secondary CPU preemption tests should pass
- Verify no regression on Pi 5 (boot_test --count 10)

### Estimated Effort
1-2 weeks

### Design Document
`docs/pi5-secondary-cpu-preemption.md` — exhaustive investigation history and design options.

johnjezl:
## Summary
- Deferred-scheduling path for Pi 5: `scheduler_tick` sets a per-CPU `reschedule_pending` flag; `el1_irq` checks it and rewrites the trap-frame ELR to `resched_trampoline`, which calls `schedule()` in task context and eret's back to the interrupted PC. See `docs/pi5-secondary-cpu-preemption.md` (Option A).
- `switch_to` (context.S) restores DAIF **last** — closes the ac46e40 race where an early DAIF restore into `DAIF.I=0` could let a timer IRQ fire mid-register-restore.
- Kill switch: CMake option `PI5_SECONDARY_PREEMPT` (default OFF).

## Status

The trampoline is correct and compiles clean, but **currently inert**: a pre-existing Pi 5 bug blocks timer IRQs from being delivered to the CPU at all (no exception vector is ever entered, despite `DAIF.I=0` and GIC `HPPIR=30`). Bisected extensively in this session:
- The ac46e40 documented fix (unmask in task_entry_trampoline + remove boot-stack daifclr) still hangs.
- Even with trampoline-arming fully disabled via a diag flag, and direct UART writes at the top of every EL1 vector, no exception fires.

Separate issue filed for the blocker so this infrastructure can land cleanly now and activate once timer-IRQ delivery is fixed.

Fixes #57 (infrastructure only — value depends on the blocker issue).

## Test plan
- [x] `make test` — QEMU, all tests green
- [x] `make kernel PLATFORM=RASPI5` — clean build
- [x] Pi 5 with default OFF: `bench smp` COMPLETED on CPUs 1/2/3 (no regression vs main)
- [ ] Pi 5 with ON: deferred until timer-IRQ-delivery blocker is resolved

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,25 @@ set_property(CACHE PLATFORM PROPERTY STRINGS "QEMU_VIRT" "JETSON_ORIN_NANO" "RAS
 # Add platform define
 add_compile_definitions(PLATFORM_${PLATFORM}=1)
 
+# Pi 5: ELR-trampoline infrastructure for deferred scheduling from the
+# timer ISR. Calling switch_to() directly from an exception handler
+# hangs on Pi 5 hardware (abandoned exception frame); the trampoline
+# runs schedule() in task context after eret. See issue #57 and
+# docs/pi5-secondary-cpu-preemption.md.
+#
+# Default OFF because a separate pre-existing bug prevents timer IRQs
+# from firing on this Pi 5 at all (commit ac46e40: tasks run with
+# DAIF.I=1 and the documented fix to unmask them triggers a shell hang
+# whose root cause is independent of this work). Turning the option
+# ON deploys the trampoline + makes all CPUs use `daifclr + wfi` in
+# idle — useful once the timer-IRQ-delivery blocker is resolved, but
+# regresses cooperative cross-CPU dispatch until then because secondary
+# CPUs wait for timer IRQs that never arrive.
+option(PI5_SECONDARY_PREEMPT "Enable deferred-scheduling preemption on Pi 5 secondary CPUs (blocked by timer-IRQ issue)" OFF)
+if(PLATFORM STREQUAL "RASPI5" AND PI5_SECONDARY_PREEMPT)
+    add_compile_definitions(PI5_SECONDARY_PREEMPT=1)
+endif()
+
 # Enable semihosting for QEMU (used for clean test exit)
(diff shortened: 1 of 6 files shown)


johnjezl:
Addressed all three review recommendations in 1f7206e:

1. **Error handling in preempt_init():** panic with allocator diagnostics (`ncmem_used()` / `NC_MEM_SIZE`) if any `ncmem_alloc` returns NULL. Both C and asm paths unconditionally deref these pointers on every IRQ, so a NULL was unrecoverable — now fails loudly at init instead of silently at first preemption.
2. **cpu_id consistency:** added a comment explaining why the trampoline uses the inline MPIDR hack rather than calling C `cpu_id()` (pre-register-save, consistent with every other `PI5_SECONDARY_PREEMPT` site, gated to `PLATFORM=RASPI5`). Flags the spot for future multi-cluster ports.
3. **Stack-size doc:** quantified 192-byte trampoline frame against `STACK_SIZE = 64 KB` (~0.3% budget).

QEMU tests remain green; Pi 5 build clean.

Issue #211, opened 2026-04-16
Lua: slm.ai_sched_decision(task_id) to inspect per-task AI decisions


johnjezl:
## Background

Deferred from the #152 binding audit. `slm.ai_sched_stats()` landed and gives aggregate counts (decisions, fallbacks, action histogram), but there's no way to ask "what did the AI scheduler decide for *this specific* task?". Useful for a demo that shows the AI making a visible choice.

## Scope

- `slm.ai_sched_decision(task_id)` → `{core_assignment, priority_adj, preempt}` table or `nil`
- Returns `nil` when `CONFIG_AI_SCHEDULER` is off or when the task hasn't been through an AI decision yet
- Wraps `ai_decode_action` from `kernel/sched/ai/ai_types.h` using the raw action index stashed on the task struct

## Design detail

The AI policy currently writes its decisions into `ai_policy_stats::action_hist` (aggregate) but does not persist per-task decisions. To expose this properly, the task struct (or a parallel per-task map) needs a "last AI action idx" field updated by `ai_*_assign_cpu`. Small amount of plumbing but a real API change.

## Demo hook

Extends option 3 (Scheduling) of `scripts/demo_menu.lua` — show overall AI stats, then drill down on one task by id.

Issue #210, opened 2026-04-16
Lua: slm.task_migrate(task_id, target_cpu) for live migration demos


johnjezl:
## Background

Deferred from the #152 binding audit. Good demo candidate for the SMP menu option — picks a running task and shows it move to a specific CPU on the next schedule point.

## Scope

- `slm.task_migrate(task_id, target_cpu)` → bool (wraps `sched_migrate_task(task, cpu)` from `kernel/include/sched.h:140`)
- Argument validation: reject negative cpu, cpu >= cpu_count, unknown task_id
- Regression test: create a task pinned to CPU 0, migrate to CPU 1, assert `slm.tasks()[task].cpu == 1` after a `slm.yield()`

## Demo hook

Would plug into option 2 (SMP) of `scripts/demo_menu.lua` — pick one of the existing tasks, migrate it, show the CPU column change.

## Safety

`sched_migrate_task` already requires the target task is not currently running, so misuse yields a clean -1. No extra guardrails needed beyond argument validation.

Issue #209, opened 2026-04-16
Lua: slm.model_load(path, name) generic ONNX loader from VFS


johnjezl:
## Background

Deferred from the #152 binding audit. Today Lua can only load the built-in MNIST model via `slm.model_load_mnist()`. The shell's `model load <path>` command can load arbitrary ONNX files from VFS but has no Lua equivalent, so scripts that want to exercise the eviction demo with different models have to go through `slm.shell_exec('model load ...')`.

## Scope

- `slm.model_load(path, name)` → model index (≥0 on success, -1 on failure)
- Read the file via `vfs_read_path`, allocate a temporary PMM buffer for the ONNX bytes, call `rust_model_load(name, data, len)`, free the PMM buffer
- Path handling should mirror the `model load` shell command (`kernel/src/shell_exec.c cmd_model`) — resolve relative to `shell_cwd` via `shell_resolve_path`

## Design detail

The ONNX buffer lifetime is: (1) allocate PMM pages for the file bytes, (2) read file into those pages, (3) hand the pointer to `rust_model_load` which copies into its own registry storage, (4) free the PMM buffer. Same pattern as the shell command — the binding should just factor the common logic into a helper if possible.

## Regression tests

- Load a valid ONNX file, assert model registry grew by one entry
- Load a non-existent path → returns -1, does not crash
- Load a non-ONNX file → returns -1, registry unchanged

Issue #208, opened 2026-04-16
Lua: slm.task_create / task_kill / task_pin for Lua-defined tasks


johnjezl:
## Background

Deferred from the #152 binding audit. Today scripts can introspect tasks via `slm.tasks()` but cannot spawn one, terminate one, or pin one to a CPU. Components and ELF programs are the only way to run work outside the shell task.

## Design challenge

A Lua-defined task needs a Lua execution context. Options:

1. **Per-task `lua_State`**: spawn the kernel task, allocate a fresh `lua_State`, serialize the function argument via `string.dump()` into bytecode, load+run the bytecode in the new state. Isolates per-task state cleanly but has a heap cost (each state is tens of KB today) and no shared upvalues — the bytecode can only use globals and its arguments.

2. **Bytecode trampoline in the shared shell state**: the new kernel task calls back into a queue consumed by the shell's `lua_State`. Work runs on the shell task's stack, not truly concurrent with other Lua code. Defeats the point of `task_create` if the goal is parallelism.

Option 1 is the real answer but doubles the Lua heap footprint per task, so this is low-priority until there's a demo that genuinely needs multiple Lua tasks.

## Scope (when implemented)

- `slm.task_create(name, fn)` → task id (fn serialized via `string.dump`)
- `slm.task_kill(id)` → bool (wraps `scheduler_remove_task` + `task_destroy`)
- `slm.task_set_priority(id, prio)` → bool
- `slm.task_pin(id, cpu)` → bool (wraps `sched_set_task_affinity`)
- Doc + help update + regression tests

## Out of scope

Lua coroutines across tasks, per-task message queues (that's `msg_subscribe` territory — #207).

PR #217, opened 2026-04-16
Lua: 9 new bindings — msg_subscribe, task_*, model_load, ai_sched_decision

johnjezl:
## Summary

Follow-ups from the [#152 audit](https://github.com/SLM-OS/SLM-Operating-System/issues/152). Closes all five tickets filed at the end of that PR.

### What lands

| Issue | Binding(s) | Kind |
|---|---|---|
| #210 | `slm.task_migrate(id, cpu)` | Wraps `sched_migrate_task` |
| #211 | `slm.ai_sched_decision(id)` | New `struct task.last_ai_action` field + decoder |
| #209 | `slm.model_load(path [,name])` | Mirrors `model load <path>` shell command |
| #207 | `slm.msg_subscribe(topic, fn)` + `msg_unsubscribe(h)` + `msg_drain()` | Callback queue drained at yield points |
| #208 | `slm.task_create(name, fn)` + `task_kill(id)` + `task_set_priority(id,p)` + `task_pin(id,cpu)` | Bytecode-trampoline, heap lock, state count |

All feature-gated bindings return `nil`/`false` when the backing feature isn't compiled in, so scripts don't need `#ifdef` equivalents.

### Infrastructure changes

- **`struct task.last_ai_action`** added inside the existing `#ifdef CONFIG_AI_SCHEDULER` block — does not shift `TASK_CONTEXT_OFFSET` for non-AI builds. `ai_assign_cpu_common` records the packed action index; `scheduler_add_task` initializes to `-1`.
- **`kernel/src/lua_stubs.c` heap lock** — `malloc`/`free`/`realloc` serialized by `spin_lock_irqsave`. Before this PR only the shell task mutated the Lua heap; Lua-defined tasks (#208) mutate from any CPU so the free list needed serialization.
- **`lua_state_count`** in `lua_slm.c` — `heap_reset` in `lua_slm_close` now only fires when the last state is closed. Resetting while another Lua task still holds allocations would be a heap UAF.

### Non-obvious fixes discovered along the way

- **`msg_router_ack` silently drops indices ≥ 32** — the original msg_router hard-codes `MAX_COMPONENTS = 32` in its `component_idx` range check. Using a naive large sentinel like `1000` would make every message re-deliver on each drain. Sentinel is `31` (top of the valid range, above the native 16-slot component region).
- **`msg_router_subscribe` is NOT idempotent** — repeated subscribes for the same `(topic, component_idx)` add a new subscriber slot with its own mailbox. Dedup at the Lua layer so duplicate `slm.msg_subscribe` calls to the same topic don't produce duplicate deliveries.
- **`sched_diag_tick` extern mismatch** (pre-existing regression caught by the new `slm.cpu_info` test-coverage in the earlier PR) — leaving alone here since it's on main now.

### Tests

19 new regression tests in `kernel/tests/test_lua.c`:

- `test_slm_task_migrate_bad_args` / `_succeeds`
- `test_slm_ai_sched_decision` / `_bad_arg`
- `test_slm_model_load_bad_paths` / `_non_onnx`
- `test_slm_msg_subscribe_basic` / `_wildcard` / `_error_isolation` / `_bad_args`
- `test_slm_task_create_basic` / `_bad_args` / `test_slm_task_lifecycle_bindings`
- Contract/arg-validation tests for each new binding (types, out-of-range, invalid ids)

`test_slm_task_create_basic` is the end-to-end test: a Lua-defined task publishes to a topic the parent subscribes to, and the parent's drain loop fires the callback — exercises bytecode round-trip, fresh-state setup, cross-state delivery via `msg_router`.

### Help + docs

- `kernel/src/help.c` — new binding rows grouped by area in the `lua` help output.
- `docs/lua.md` — full rows with return-type contracts and feature-gating notes.
- `kernel/include/lua_slm.h` — summary block lists every new binding with its `#` reference.

## Test plan

All builds pass on all four platforms:
- [x] `make kernel` (QEMU_VIRT ARM64)
- [x] `make kernel PLATFORM=X86_64`
- [x] `make kernel PLATFORM=RASPI5`
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO`

All tests pass across all relevant configurations:
- [x] `make test WORK_STEALING=OFF` — PASSED (3/3 consecutive runs)
- [x] `make test WORK_STEALING=OFF AI_SCHED=ON AI_EVICTION=ON` — PASSED

Default `WORK_STEALING=ON` on QEMU still shows pre-existing SMP flakiness documented in CLAUDE.md — unrelated to this PR.


johnjezl:
Addressed the three review items in `8eb5e53`:

- **Critical (AI decode duplication)**: lua_slm.c now `#include`s `ai_types.h` (plain stdint + static_asserts, already on KERNEL_INCLUDES) and calls `ai_decode_action()` via a `struct ai_sched_action`. If the AI action encoding ever changes, `sched_ai.c` and `lua_slm.c` update together — no more silent skew.
- **Warning (drain batch)**: Raised the per-drain cap from 64 to 256 and gave it a name (`LUA_MSG_DRAIN_BATCH`) with a comment documenting that messages are *not* lost — the router holds the next one for the next drain cycle.
- **Warning (sentinel collision)**: Added two `_Static_assert`s: `LUA_MSG_SUB_IDX > COMPONENT_MAX_COUNT` (traps a future native-component bump) and `LUA_MSG_SUB_IDX < 32` (locks the `msg_router_ack` upper bound). The comment points at `runtime/src/msg_router.rs:MAX_COMPONENTS` if either fires.

`make test WORK_STEALING=OFF` and `make test WORK_STEALING=OFF AI_SCHED=ON AI_EVICTION=ON` both PASS. All four platforms (QEMU, X86_64, RASPI5, JETSON_ORIN_NANO) build cleanly with `AI_SCHED=ON`.

Issue #180, opened 2026-04-15
Investigate Theory Radar for symbolic page-eviction and scheduler policies


johnjezl:
## Platform
All — applies to any build with `AI_SCHED=ON` or the `ai_eviction` runtime feature. The artifact (a discovered formula) is architecture-independent.

## Subsystem
`runtime/mm/eviction` (XGBoost / int8 MLP / CACHEUS policies) and `kernel/sched` + AI scheduler. Tooling work would live in the sibling `slm-os-page-sim` reference crate, not in the kernel.

## Summary
Investigate whether **Theory Radar** (https://github.com/ahb-sjsu/theory-radar) — an SJSU symbolic formula-discovery tool — can produce small interpretable formulas that match or approximate the trained ML eviction and scheduling policies in SLM-OS, eliminating the runtime cost of carrying neural-network weights on-device.

## Background — What is Theory Radar?

Theory Radar answers the question *"do you need a black box, or would a formula suffice?"* It exhaustively (then with beam-search + meta-learned pruning) searches over symbolic formula trees built from the input features, evaluating each candidate via threshold-tuned F1 against labels. The output is a `FormulaTrace` like `min(insulin, age) + glucose` plus the threshold and statistical significance.

**Three-tier architecture:**
1. Core engine — exhaustive enumeration with exact F1 via threshold optimization, leveraging a Monotone Invariance Theorem to reuse evaluations across monotone transforms
2. Search accelerators — beam search with cost-plus-heuristic ordering, meta-learned pruning that drops 88-99.6% of dead subtrees with no false negatives
3. Representation augmenters — PCA, PLS, Tucker, kernel projections so a depth-3 formula can implicitly access many features

**Inputs:** tabular `(X, y)` with binary labels.
**Outputs:** a small symbolic expression + threshold + fairness-tuned test metrics.

## Current Hypothesis

The eviction subsystem already has a 27-feature vector (`runtime/src/mm/eviction/features.rs:1-80`) that is the *exact* input to the trained int8 MLP and XGBoost policies (`mm/eviction/mlp.rs`, `mm/eviction/xgboost.rs`, generated weights under `mm/eviction/generated/`). Training data for these policies was produced by the sibling `slm-os-page-sim` simulator with a known oracle (which candidate would have been the best to evict). That gives Theory Radar exactly what it needs:

- **X** = the 27-feature rows the simulator already logs per decision
[...]

Issue #179, opened 2026-04-15
mathf::tanhf: add inline comment justifying ±4 saturation boundary


johnjezl:
## Summary

\`mathf::tanhf\` in \`runtime/src/inference/mathf.rs\` saturates
to ±1 past x = ±4. The choice is correct (Padé(7,7)'s numerator
outgrows the denominator past ±4 and returns values > 1.0 —
mathematically impossible for tanh), but the inline code comment
just says "saturates at ±1 outside [-4, 4]" without spelling out
the why.

The module-level docstring covers it, but a future reader looking
at the function in isolation might wonder if ±5 (the previous
window) or ±6 would be safer.

## Suggested

Expand the function's preamble comment to include a one-liner
explaining the math:

\`\`\`rust
/// Hyperbolic tangent via a Padé(7,7) rational approximation.
///
/// Saturates at ±1 past x = ±4 because the Padé numerator
/// (degree 7 in x) outgrows the denominator past ±4 and the
/// rational returns values > 1.0 — mathematically impossible
/// for tanh (|tanh(x)| < 1 ∀ x ∈ ℝ). Inside [-3, 3] the
/// approximation is < 1e-6 off; the GELU caller's tanh
/// argument lands well inside that range in practice.
\`\`\`

(Most of this is already in the module-level docstring; the ask
is to put the *one-line justification* of the boundary choice
right at the function so callers don't have to scroll.)

## Acceptance

- A reader looking at \`mathf::tanhf\` in isolation can answer
  \"why ±4?\" from the inline comment.

## Context

Discovered during PR #173 review.

johnjezl:
Closed via PR #189 (squash merge 5b69bd1).

johnjezl:
## Blocker anticipated from PR #333 review

The translator's `default:` case now returns `HAILO_ERR_INVAL` for any oneof tag it has no translator for (PR #333, third-pass review). This is correct — silent drops produce semantically-wrong streams — but it means real HEFs with branches we don't yet translate (`write_data_ccw` / `write_data` / etc. in compute contexts) will fail the load, not just emit a degraded stream.

Before wiring the translator into `load_model`:

[...]

Issue #178, opened 2026-04-15
x86-multitask-boot-test.sh: hardcoded socket timeouts may false-fail under system load


johnjezl:
## Summary

\`scripts/tests/x86-multitask-boot-test.sh\` (the P1-2 reproducer)
hardcodes ser2net socket timeouts in the embedded Python multi-
task driver:

- Connect timeout: 30 s
- Per-receive timeout: 8 s (in \`drain()\`)
- Per-loop \`drain()\` deadline: 4 s

These are generous enough on a quiet dev machine but could
false-fail when:

- The dev box is under heavy CI load (other QEMU runs, Cargo
  builds, etc.) and the ser2net daemon's read-ready notification
  is delayed.
- Network jitter — even on localhost — pushes a small percentage
  of recv() calls past the 4 s deadline.
- The shell on test-pc takes longer than usual to print
  \`slmos>\` after a command (e.g. when running on a pre-warm
  scheduler).

The boot-reliability gate (\`labctl boot-test\`) already has a
configurable \`--timeout\` and defaults to 45 s; the multi-task
loop should match.

## Suggested

- Promote the timeouts to env-var-overridable defaults:
  \`SLMOS_RPC_DRAIN_TIMEOUT\`, \`SLMOS_RPC_PROMPT_TIMEOUT\`.
- Bump the prompt-wait deadline to match the boot-test default
  (10 s for in-shell commands, 45 s for boot).
- On false-fail, retry once with 2× timeout before declaring
  failure — distinguishes "really stuck" from "scheduler
  hiccup".

## Acceptance

- Script tolerates a 2× CPU steal from a concurrent build
  without false-failing.
- Genuine SLM-OS hangs still surface as failures within 60 s
  per iteration.

## Context

Discovered during PR #173 review. Hasn't false-failed in
practice (5/5 ran clean in the dated 2026-04-15 results), but
worth hardening before this script joins CI.

PR #335, opened 2026-04-19
Pi 5 Hailo-8 Phase 6.5 (#178): boundary channels + OpenBoundary in ACTIVATION

johnjezl:
## Summary

Closes (most of) #178. Extends the context-switch translator so ACTIVATION carries OpenBoundaryInput/Output actions for each boundary edge in the HEF, alongside the existing BURST_CREDITS_TASK_RESET. This is the concrete test of the hypothesis driving #180: that fw v4.23 rejects subsequent CORE-CPU RPCs with 0xFFFFFFFF because ACTIVATION wasn't carrying enough content for firmware to consider the context "complete."

- **`hailo_cs_translate_cfg` gains** four boundary descriptor-list fields (input IOVA + desc count, output IOVA + desc count, shared page size). Load_model (#179) is responsible for allocating the VDMA descriptor lists and populating these before calling the translator.
- **`translate_activation`** now walks `info->pads[]` and, for each pad with `has_stream_info=true`, emits OPEN_BOUNDARY_INPUT_CHANNEL (is_input) or OPEN_BOUNDARY_OUTPUT_CHANNEL. Uses the `HAILO_CS_BOUNDARY_{INPUT,OUTPUT}_CHANNEL_OFFSET` constants already shared with `translate_allow_input_dataflow` so the two call sites agree by construction.
- **Caller-mistake detection:** if the HEF has a boundary pad but the matching `cfg->boundary_*_desc_list_iova` is zero, translation fails with `HAILO_ERR_INVAL` instead of emitting an action with a NULL DMA address that firmware would interpret as garbage.
- **Tests (+6 cases):** byte-for-byte wire format for both OPEN_BOUNDARY directions (36B input, 28B output), full single-stream MLP ACTIVATION layout (72B total), internal pads correctly skipped, missing-IOVA error paths covered.

## Commit

- \`91e837c\` — Phase 6.5 (#178): boundary channels + OpenBoundaryInput/Output in ACTIVATION

## Remaining Phase 6 work (not in this PR — tracked)

- **#179** — wire the translator into \`inference_device_hailo::load_model\` (replace the best-effort CCW+CONFIG_STREAM path with the full 6-step ctxsmoke-equivalent flow, allocating boundary VDMA descriptor lists first). Load_model scope is the substantive integration work; worth its own PR.
- **#180** — hardware iteration on pi-5-1 to close the 0xFFFFFFFF CORE-CPU blocker. This PR sets up the content hypothesis; confirming it needs deploy + trace.
- **Open caveats from PR #333 review trail** flagged on #179:
  - Translator's default-case policy rejects any HEF oneof tag it doesn't translate yet (write_data_ccw in DYNAMIC contexts is the likely real-HEF surprise). Audit a compiled MobileNetV1 HEF before #179 wires up.
  - \`sequencer_config\` packed-43 vs naturally-aligned-48 risk (same lesson as common_action_header 5→8 byte surprise). Cross-check against a live firmware trace before shipping TRIGGER_SEQUENCER on real hardware.

## Test plan

- [x] \`make test AI_SCHED=ON\` — full QEMU suite green (includes the 6 new ACTIVATION cases).
- [x] \`make kernel PLATFORM=RASPI5 AI_SCHED=ON\` — clean.
- [x] \`make kernel PLATFORM=JETSON_ORIN_NANO\` — clean.
- [x] \`make kernel PLATFORM=X86_64\` — clean.
- [ ] Hardware verification on pi-5-1: once #179 wires the translator into load_model, confirm the richer ACTIVATION unblocks #180 (or, if not, narrows the hypothesis).

### DIFF
--- a/kernel/ai_accel/hailo/hailo_cs_translator.c
+++ b/kernel/ai_accel/hailo/hailo_cs_translator.c
@@ -81,14 +81,138 @@ int hailo_cs_translate_application_header(
 /* ACTIVATION context                                                           */
 /* -------------------------------------------------------------------------- */
 
-/* Minimum legal ACTIVATION per HailoRT's fill_activation_config_recepies:
- * BURST_CREDITS_TASK_RESET (zero body). Real MLP ACTIVATION contexts
- * also carry OpenBoundaryInput/Output per boundary edge layer; those
- * land once the parser surfaces edge_layer → VDMA channel mapping. */
-static int translate_activation(struct hailo_cs_builder *b)
+/* ACTIVATION context (Phase 6.5, #178).
+ *
+ * Per HailoRT's fill_activation_config_recepies (v4.23 source trail
+ * documented in docs/pi5-ai-hat-plan.md §6.3):
+ *
+ *   1. BURST_CREDITS_TASK_RESET (zero body) — resets per-channel
+ *      burst credit counters so the new network's config channel
+ *      starts clean.
+ *   2. OPEN_BOUNDARY_INPUT_CHANNEL per boundary input edge — binds
+ *      a host→device VDMA descriptor list to the channel firmware
+ *      will pull tensor data through during inference.
+ *   3. OPEN_BOUNDARY_OUTPUT_CHANNEL per boundary output edge — binds
+ *      device→host channel for the response tensor.
+ *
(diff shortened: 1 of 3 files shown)


johnjezl:
Superseded — combined with #179 into #336 per request.

Issue #253, opened 2026-04-17
Pi 5: GeeekPi AI HAT+ (Hailo-8/8L) support — PCIe driver, NPU bring-up, inference


johnjezl:
## Summary

Implement support for the **GeeekPi AI HAT+** (Hailo-8L, 13 TOPS, PCIe Gen3 x1) on the Raspberry Pi 5, end-to-end from PCIe enumeration through running an AI inference workload from an SLM-OS shell command.

**Detailed plan:** [`docs/pi5-ai-hat-plan.md`](../blob/main/docs/pi5-ai-hat-plan.md)

## Why

The capstone already has a CPU-only AI scheduler (MLP + PPO via NEON/SSE) and a portable NVIDIA GSP inference stack targeting x86-64 and Jetson. The Pi 5 has no AI acceleration path today — `docs/capstone-feature-status.md`'s GPU Inference row shows "—" for Pi 5. Adding the AI HAT+ closes that gap with a realistic, purchasable, under-$100 accelerator and gives the demo a third, visually distinct AI platform story.

## Scope

The plan is staged in 7 phases:

- **Phase 0 — Research & prerequisites** (~1 week): Hailo protocol notes, `pcie1` + MIP1 register map, `.hef` model frozen.
- **Phase 1 — ARM64 PCIe host controller** (~2–3 weeks): `kernel/drivers/pcie/pcie_bcm2712.c`, config-space API, BAR mapping, MSI-X allocator.
- **Phase 2 — Inference device abstraction** (~1 week, overlaps Phase 1): `struct inference_device_ops`, CPU backend wraps existing NEON MLP.
- **Phase 3 — Hailo probe & control plane** (~2 weeks): `kernel/ai_accel/hailo/` with core + platform shim, mirroring the NVIDIA GSP pattern.
- **Phase 4 — Firmware + model load** (~2 weeks): firmware upload via VDMA control channel, `.hef` parser, weight DMA.
- **Phase 5 — Single-model inference** (~2 weeks): submit/complete protocol, `hailo infer` shell command, latency histogram.
- **Phase 6 — AI scheduler integration** (~1–2 weeks): scheduler MLP recompiled as `.hef`, policy switch at runtime.
- **Phase 7 — Shell integration & demo** (~1 week): Lua bindings, demo script, capstone-status update.

## Capstone-realistic target

[...]

PR #343, opened 2026-04-21
#180: fix Hailo context-switch handshake — 5-byte header + drop FETCH_CCW_BURSTS in PRELIMINARY

### LINKED ISSUE #180: Investigate Theory Radar for symbolic page-eviction and scheduler policies

johnjezl:
## Summary

Resolves #180. After 13 commits of bisecting and a HailoRT v4.23 wire capture against a real Hailo-8L Model Zoo HEF, the SLM-OS Hailo context-switch handshake completes cleanly on pi-5-1:

```
[5/8] SET_CONTEXT_INFO(ACTIVATION, 63 B)       rc=0  ✅
[6/8] SET_CONTEXT_INFO(BATCH_SWITCHING, 16 B)  rc=0  ✅
[7/8] SET_CONTEXT_INFO(PRELIMINARY, 26 B)      rc=0  ✅
[8/8] SET_CONTEXT_INFO(DYNAMIC, 5 B)           rc=0  ✅
```

Two root causes, both proven by ground-truth wire capture:

### 1. `common_action_header` is 5 bytes, not 8

A prior session's analysis claimed firmware reads the header with natural alignment (1-byte action_type + 3 padding + 4-byte time_stamp). That was wrong — `#pragma pack(1)` IS honored. HailoRT emits `BURST_CREDITS_TASK_RESET` (a zero-body action) as exactly `1e ff ff ff ff` before the next action begins. Also `time_stamp` is `0xFFFFFFFF` (`CONTEXT_SWITCH_DEFS__TIMESTAMP_INIT_VALUE`), not 0.

Fix: `kernel/ai_accel/hailo/hailo_cs_actions.h` — struct now `__attribute__((packed))`, `_Static_assert` is 5 bytes, `HAILO_CS_TIMESTAMP_INIT_VALUE` define added, builder writes the INIT value into every header.

### 2. `FETCH_CCW_BURSTS` is not supported in PRELIMINARY on Hailo-8L

Wire capture confirms HailoRT does not emit `FETCH_CCW_BURSTS` (action_type 27) directly in PRELIMINARY for Hailo-8L. Firmware rejects it with `0x402a0001 = CONFIG_MANAGER_WRAPPER_STATUS_ACTION_TYPE_NOT_SUPPORTED`. HailoRT uses a `REPEATED_ACTION`-wrapped `AddCcwBurst` path instead.

Fix: `translate_preliminary` now emits `ACTIVATE_CFG_CHANNEL` alone. Real CCW loading via the wrapped path is a separate workstream (Phase 6.10).

### Smaller correctness fixes landed alongside

Each was hardware-verified to be necessary even after the header fix:

- `host_buffer_info.bytes_in_pattern = pad->core_bytes_per_buffer` (matches `vdma_edge_layer.cpp:73`); was hardcoded 0.
- `HAILO_CS_BOUNDARY_OUTPUT_CHANNEL_OFFSET` 2 → 15 so OUTPUT lands at channel 16 (first valid D2H per HailoRT `channel_allocator.cpp` MIN_D2H=16/MAX_D2H=31).
- Translator emits OUTPUT before INPUT in ACTIVATION (matches `resource_manager_builder.cpp:1059-1075`).

### How the wire bytes were captured

1. Swap pi-5-1's Pi OS SD card into the SDWire (memory note `pi5_lab_setup.md` covers the dual-card workflow).
2. Load the instrumented driver:
   ```
   sudo rmmod hailo_pci
   sudo insmod /home/pi/Downloads/hailort-drivers/linux/pcie/hailo_pci.ko
   ```
   `hailo_pcie_write_firmware_control` gets `print_hex_dump` on every FW-control RPC.
3. Patch `libhailort.so.4.23.0` to bypass the hardcoded `max_desc_page_size = 4096` guard (Hailo-8L Model Zoo HEFs require 16384):
   ```
   sudo cp /usr/lib/libhailort.so.4.23.0 /usr/lib/libhailort.so.4.23.0.bak
   for off in 0x24a0cc 0x24a8f4 0x24a998; do
     printf '\x13' | sudo dd of=/usr/lib/libhailort.so.4.23.0 \
       bs=1 count=1 seek=$((off+1)) conv=notrunc
   done
   ```
   Three 1-byte ARM64 patches change `cmp wN, #0x1000` to `cmp wN, #0x4000` inside `BufferSizesRequirements::get_buffer_requirements_multiple_transfers`.
4. Run a Hailo-8L Model Zoo HEF (`mobilenet_v1.hef`) and capture dmesg.
5. Restore the original libhailort.

Resulting capture cached at `docs/reference/hailort-v4.23.0-wire-capture-mobilenet.txt` for future sessions.

## Test coverage

Every fix has a regression test that would catch it if it came back:

| Change | Test |
|---|---|
| 5-byte packed header | `test_cs_builder_append_emits_header_then_body`, `test_cs_builder_appends_concatenate`, `test_cs_builder_returns_nomem_on_overflow`, `test_cs_builder_accepts_zero_body_action` |
| `time_stamp = HAILO_CS_TIMESTAMP_INIT_VALUE` | `test_cs_builder_append_emits_header_then_body` (asserts the constant directly), `test_cs_builder_accepts_zero_body_action` (asserts `0xFF` bytes), `test_cs_translate_activation_emits_open_boundary_input/output` (asserts INIT bytes inside translator output) |
| `bytes_in_pattern = core_bytes_per_buffer` | `test_cs_translate_activation_emits_open_boundary_input` and `_output` (new explicit assertions at body offset +26) |
| `OUTPUT_OFFSET = 15` (D2H range) | `test_cs_translate_application_header_boundary_bitmap` (bit 16), `test_cs_translate_activation_emits_open_boundary_output` (packed=0x10), `test_cs_translate_activation_emits_input_and_output` (same), `test_inf_hailo_run_reuses_load_boundary_iovas` (channel 16 register location) |
| OUTPUT-before-INPUT ordering | `test_cs_translate_activation_emits_input_and_output` (asserts OUT body at [5..29], IN at [30..62]) |
| Dropped FETCH_CCW_BURSTS in PRELIMINARY | `test_cs_translate_contexts_produces_all_four` (PRELIMINARY length = 26) |

`make test` passes (full QEMU suite). Pi 5 build clean. Hardware-verified on pi-5-1 fw v4.23.

## Cached HailoRT v4.23 reference sources

`docs/reference/` gains seven cached HailoRT v4.23.0 files used during the investigation, so future sessions can consult them without re-fetching:

- `hailort-v4.23.0-channel_allocator.cpp` — MIN/MAX_H2D/D2H constants
- `hailort-v4.23.0-channel_id.hpp` — ChannelId struct
- `hailort-v4.23.0-hailort_driver.hpp` — full driver header with channel index ranges
- `hailort-v4.23.0-vdma_edge_layer.cpp` + `.hpp` — `bytes_in_pattern = transfer_size` ground truth
- `hailort-v4.23.0-get_boundary_buffer_info.md` — call-chain summary
- `hailort-v4.23.0-wire-capture-mobilenet.txt` — full capture (574 lines: every FW-control RPC for one inference)

## Documentation updates

- `docs/pi5-ai-hat-plan.md` Phase 6.9 section rewritten with the resolution writeup, both root causes, the reproducible wire-capture procedure, and a forward pointer to Phase 6.10 (real CCW loading).
- The "Key finding — header is 8 bytes" section now carries an explicit retraction.
- The hardware-iteration table footnotes that the "8-byte → rc=0" row was a coincidence.
- Test-list updates reflecting the dropped FETCH_CCW_BURSTS tests.

## Out of scope

- **Real CCW loading via REPEATED_ACTION + AddCcwBurst sub-actions.** SLM-OS still doesn't actually transfer weights. Phase 6.10 work.
- **Inference dispatch** (`CHANGE_CONTEXT_SWITCH_STATUS(ENABLED)` + per-frame VDMA submission). Blocked on Phase 6.10.
- **Removing diagnostic scaffolding from ctxsmoke.** The hexdump / IOVA print / BAR4 scope scan / 500ms delay are kept as a permanent probe until the full Hailo bringup is past Phase 6.10.

## Test plan
- [x] `make test` (QEMU): all tests pass
- [x] `make kernel PLATFORM=RASPI5 HAILO_FW_BLOB=…`: clean build
- [x] Hardware verification on pi-5-1 fw v4.23 with `hailo ctxsmoke`: all 4 contexts rc=0

### DIFF
--- a/docs/pi5-ai-hat-plan.md
+++ b/docs/pi5-ai-hat-plan.md
@@ -459,7 +459,7 @@ The natural continuation of 6.3. HEF proto stores structured `ProtoHEFAction` on
 **`kernel/ai_accel/hailo/hailo_cs_actions.h`** — host-side mirrors of the wire structs:
 
 - `enum hailo_cs_action_type` — all 45 action-type values (v4.23 order).
-- `struct hailo_cs_common_action_header` — **8 bytes on the wire** (1-byte action_type + 3 pad + 4-byte time_stamp). This is a firmware-discovered fact; see "Key finding" below.
+- `struct hailo_cs_common_action_header` — **5 bytes on the wire** (1-byte action_type + 4-byte time_stamp, packed). `time_stamp` is set to `HAILO_CS_TIMESTAMP_INIT_VALUE` (`0xFFFFFFFF`) for every action. *(Earlier this was incorrectly believed to be 8 bytes with natural alignment; see Phase 6.9 wire-capture writeup for how that was overturned.)*
 - `struct hailo_cs_host_buffer_info` (19 B) — embedded in `ACTIVATE_*` actions for firmware to DMA-pull from host buffers.
 - `struct hailo_cs_stream_reg_info` (19 B) — NN-stream buffer geometry.
 - Preliminary-context bodies: `ACTIVATE_CFG_CHANNEL` (21 B), `FETCH_CCW_BURSTS` (3 B), `DEACTIVATE_CFG_CHANNEL` (2 B).
@@ -480,13 +480,29 @@ hailo_control_set_context_info(HAILO_CS_CONTEXT_TYPE_PRELIMINARY,
 
 **`hailo ctxsmoke` shell command** — exercises the full 6-step bring-up chain (RESET → SET_NETWORK_GROUP_HEADER → 4 SET_CONTEXT_INFO calls) against live firmware with per-context minimum action stubs. Used for hardware-level wire-format validation; kept as a permanent diagnostic probe.
 
-#### Key finding — `common_action_header_t` is **8 bytes**, not 5
+#### Key finding (RETRACTED 2026-04-20) — `common_action_header_t` is 5 bytes, not 8
 
-Firmware v4.23 compiles `CONTEXT_SWITCH_DEFS__common_action_header_t` with natural alignment despite the `#pragma pack(push, 1)` wrapping the reference header. The packed u8 `action_type` is followed by 3 pad bytes, then the u32 `time_stamp` — total 8 bytes on the wire. 5-byte headers produce `major=0x40130016` = `CONTEXT_SWITCH_STATUS_MISALIGNMENT_ERROR_WHILE_READING_ACTIONS`; 8-byte headers are accepted.
+The original 6.4d writeup claimed firmware v4.23 reads
+`CONTEXT_SWITCH_DEFS__common_action_header_t` with natural alignment
+(8 bytes) despite the `#pragma pack(1)` in the reference. **That was
+wrong.** A HailoRT v4.23 wire capture against a real Hailo-8L Model
+Zoo HEF on 2026-04-20 (cached at
+`docs/reference/hailort-v4.23.0-wire-capture-mobilenet.txt`) shows
+`BURST_CREDITS_TASK_RESET` — a zero-body action — emitted as exactly
+5 bytes: `1e ff ff ff ff` before the next action header begins.
 
-Root cause is firmware-internal (likely compiled without the pragma visible, or a wrapper struct re-imposes alignment). Takeaway: **cross-check wire struct sizes against hardware, don't trust the pragma**. Body structs (`host_buffer_info_t`, `activate_cfg_channel_t`, etc.) still appear to be 1-byte packed.
+The earlier MISALIGNMENT_ERROR_WHILE_READING_ACTIONS observation that
+seemed to validate the 8-byte theory was actually a *consequence* of
+something else in the action stream (likely the
+`bytes_in_pattern=0` and `OUTPUT_OFFSET=2` bugs we also carried at
+that time). Once the header was packed back to 5 bytes AND the other
+fixes landed, all 4 contexts return rc=0. See Phase 6.9 below for
+the full root-cause writeup. Body structs are also 1-byte packed,
+which the pragma does correctly enforce.
 
 #### Hardware iteration on pi-5-1 fw v4.23 (2026-04-19)
+
+*Note: the bottom row of this table — "8-byte common header → ACTIVATION rc=0" — was a coincidence: the rc=0 came from a different change in the same iteration (likely action-type ordering or body content), not the header size. The actual header is 5 bytes; see Phase 6.9 retraction.*
 
 Each iteration consumed one firmware rejection code to decode the next layer:
 
@@ -586,9 +602,8 @@ HEF parser EnableLcu extraction, 6.4g (3):
 Translator application_header + contexts, 6.4f (6):
 - `test_cs_translate_application_header_fills_defaults` — verifies every derived field in the 32-byte header.
 - `test_cs_translate_application_header_rejects_null` — null-arg.
-- `test_cs_translate_contexts_produces_all_four` — byte-for-byte assertion of all four context streams (lengths 8/16/40/8, action_type bytes at expected offsets, `host_buffer_info.dma_address` round-trip).
-- `test_cs_translate_contexts_uses_ccw_count_for_burst_count` — `info.ccw_action_count=7` → `FETCH_CCW_BURSTS.ccw_bursts=7`.
-- `test_cs_translate_contexts_clamps_burst_count_to_u16` — 100000 actions → clamped to UINT16_MAX.
+- `test_cs_translate_contexts_produces_all_four` — byte-for-byte assertion of all four context streams (lengths 5/10/26/5 with the 5-byte header + dropped FETCH_CCW_BURSTS, action_type bytes at expected offsets, `host_buffer_info.dma_address` round-trip). *(Phase 6.9 update — was 8/16/40/8 before the header retraction and dropped FETCH_CCW_BURSTS.)*
+- *(Removed in Phase 6.9: `test_cs_translate_contexts_uses_ccw_count_for_burst_count` and `test_cs_translate_contexts_clamps_burst_count_to_u16` — both asserted FETCH_CCW_BURSTS in PRELIMINARY, which is no longer emitted.)*
 - `test_cs_translate_contexts_rejects_null` — null-arg.
 
 Translator EnableLcu, 6.4g (3):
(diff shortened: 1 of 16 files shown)


johnjezl:
Note: the "#180" shorthand used throughout this PR's commits and docs was a memory-note mistake — it does not correspond to GitHub issue #180 (which is about Theory Radar page-eviction). That issue was auto-closed by the merge and has been reopened. The actual Hailo umbrella issue is #253 (Pi 5 AI HAT+ support). No functional impact; just a labeling error to be aware of when reading the commit history.

Issue #372, opened 2026-04-25
Jetson GA10B: replace 500 ms post-sentinel sleep with SEMAPHORE_RELEASE-on-completion


johnjezl:
## Background

`scripts/gpu-launch-common.c:gpu_submit_and_poll()` currently sleeps 500 ms after each per-op sentinel match as a workaround for the multi-CTA timing race introduced during M2 (parameterized GEMM) work in PR #364:

> A fast CTA can write the sentinel before lagging CTAs finish their writes,
> leaving spurious zeros in those cells.

For an 8-op MNIST pipeline that's **4 seconds of pure sleeps** added to the launcher's pre-kexec setup. The user-facing post-kexec dispatch path (M7's `slm.gpu_run_mnist()`, kernel-side `ga10b_submit_and_poll`) doesn't go through this sleep — it uses busy-wait polling — so the runtime experience is fast, but the launcher setup cost is meaningful and the workaround is fragile (CTAs whose tail latency exceeds 500 ms would still desync).

## Proper fix

Append a `SEMAPHORE_RELEASE` method to the dispatch pushbuffer (the same family already used by `ga10b_build_compute_sema_release_pushbuffer` for Phase 7). The semaphore writes the configured payload to a configured GPU VA only **after** all SMs have drained, so the launcher polls that location for exact match — no sleep needed, no multi-CTA race.

Ampere methods (`AMPERE_COMPUTE_B = 0xC7C0`):
- `NVC7C0_SET_REPORT_SEMAPHORE_PAYLOAD_LOWER` (0x0158)
- `NVC7C0_SET_REPORT_SEMAPHORE_PAYLOAD_UPPER` (0x015C)
- `NVC7C0_SET_REPORT_SEMAPHORE_ADDRESS_LOWER` (0x0160)
- `NVC7C0_SET_REPORT_SEMAPHORE_ADDRESS_UPPER` (0x0164)
- `NVC7C0_REPORT_SEMAPHORE_EXECUTE` (0x0168) — `OP=RELEASE | STRUCTURE_SIZE=ONE_WORD`

## Scope (rough)

- Wire-format change: extend `struct ga10b_pipeline_op` with `sem_gpu_va` (8 B). Bumps sizeof 24 → 32, page capacity 170 → 128. v5 → v6 handoff.
[...]

PR #373, opened 2026-04-25
Jetson GA10B M7: Lua + shell entrypoints for GPU MNIST inference

johnjezl:
## Summary

The deferred M7 milestone from `docs/jetson-gpu-mnist-plan.md`: end-to-end MNIST inference reachable from Lua and the shell, post-kexec. After PR #364 landed M0–M8 of the GPU compute path, the only remaining gap was a clean entrypoint — running the pipeline took three sequential `nvgpu` shell commands plus manual `peek` reads. M7 wraps it:

```
lua-admin> local bytes, argmax = slm.gpu_run_mnist()
        → 3      (the predicted MNIST class)
```

```
slmos> nvgpu run-mnist
        → dispatches the v5 pipeline + reads logits + prints argmax
```

The Lua API lazy-initialises the bringup state on first call (walks `nvgpu inherit` + `nvgpu channel` automatically), so end-to-end MNIST inference is reachable from a single Lua line post-kexec.

## What shipped

**Kernel-side helpers (`kernel/gpu/nvidia/ga10b_bringup.{h,c}`)**:
- `ga10b_bringup_read_pipeline_output(b, out, cap)` — reads `cap` bytes from the LAST v5 pipeline op's `output_phys`, with cache-invalidate matching the submit-and-poll pattern.

**FFI (`kernel/include/slm_ffi.h` + `kernel/src/slm_ffi.c`)**:
- `slm_gpu_run_mnist(void *logits_bytes_out)` — lazy-initialises a persistent `struct ga10b_bringup`, dispatches the pipeline, copies 40 bytes of fp32 bit patterns into the caller's buffer. Non-Jetson builds stub it to `-1`.
- `slm_fp32_argmax(const void *logits, uint32_t n)` — FP-free argmax over fp32 bit patterns. Required because the kernel target compiles with `-mgeneral-regs-only` (no fp32 in C). Bit-pattern math handles same-sign and different-sign cases without touching FP registers.

**Shell (`kernel/src/shell_sys.c`)**:
- `nvgpu run-mnist` subcommand: calls launch_kernel + reads logits + prints them as raw fp32 bit patterns + argmax.

**Lua (`kernel/src/lua_slm.c`)**:
- `slm.gpu_run_mnist()` admin function: returns `(40-byte string, argmax)` on success, `(nil, rc)` on failure. Float decoding happens Lua-side via `string.unpack("<f", bytes, 1 + i*4)` since `string.unpack` lives in `liblua` (compiled with FP support).

## Hardware validation (jetson-nano-1, 2026-04-25)

Pre-flight: `scripts/gpu-kernel-mnist.c --preserve-for-kexec` populated the v5 handoff. Then kexec into SLM-OS.

**Shell command:**
```
slmos> nvgpu inherit
slmos> nvgpu channel
slmos> nvgpu run-mnist
[GA10B-P8] pipeline mode — 8 ops, ops_phys=0x15a3a6000
... 8 ops fire ...
[GA10B-P8] pipeline complete — 8 ops fired
run-mnist: predicted class = 3
```

**Lua API** (single call — no manual inherit + channel):
```
slmos> lua-admin -e print(select(2,slm.gpu_run_mnist()))
[GA10B-P8] pipeline mode — 8 ops, ops_phys=0x15a3a6000
... pipeline runs ...
3
```

GP_PUT advanced 8 → 24 across the two runs (8 dispatches each), confirming the lazy-init reuses the channel across calls. Both paths produce **argmax = 3 = the CPU NEON reference's argmax** for the same synthetic input.

## Test plan

- [x] Host tests: `make test-ga10b-bringup` — 74/74 pass (no new tests; existing predicate tests still cover the v5 path)
- [x] Jetson kernel build: `make kernel PLATFORM=JETSON_ORIN_NANO` clean
- [x] QEMU ARM64 build: `make kernel` clean (non-Jetson stub returns -1)
- [x] Hardware validation: shell + Lua paths both produce argmax = 3 on jetson-nano-1
- [x] Lazy-init across multiple Lua calls validated (GP_PUT advances cumulatively across runs)

## Out of scope (tracked separately)

The 500 ms post-sentinel sleep in `gpu_submit_and_poll` (Linux launcher) is unchanged. That sleep is a one-time pre-kexec setup cost; M7's user path goes through SLM-OS post-kexec which busy-waits and doesn't sleep. The proper fix (SEMAPHORE_RELEASE-on-completion) is a wire-format change — filed as **#372**.

### DIFF
--- a/kernel/gpu/nvidia/ga10b_bringup.c
+++ b/kernel/gpu/nvidia/ga10b_bringup.c
@@ -1731,3 +1731,38 @@ int ga10b_bringup_run(struct ga10b_bringup *b)
     uart_puts("[GA10B] bringup complete — channel open, method accepted\n");
     return 0;
 }
+
+int ga10b_bringup_read_pipeline_output(struct ga10b_bringup *b,
+                                        void *out, size_t cap)
+{
+    if (!b || !out) return -1;
+    if (g_handoff.version < 5u || g_handoff.pipeline_n_ops == 0u) {
+        return -1;
+    }
+    if (g_handoff.pipeline_ops_phys == 0u ||
+        g_handoff.pipeline_n_ops > GA10B_PIPELINE_MAX_OPS) {
+        return -1;
+    }
+
+    /* Resolve the LAST op's output_phys. Same identity-DRAM-mapping
+     * assumption as the pipeline runner — physical address read from
+     * DRAM is also a valid VA SLM-OS can dereference at EL2. */
+    const struct ga10b_pipeline_op *ops =
+        (const struct ga10b_pipeline_op *)
+            (uintptr_t)g_handoff.pipeline_ops_phys;
+    const struct ga10b_pipeline_op *last =
+        &ops[g_handoff.pipeline_n_ops - 1u];
+    if (last->output_phys == 0u) return -1;
+
+    /* Invalidate the buffer's cache range before the read. The GPU
+     * wrote the data via its own (uncached-from-CPU's-perspective)
+     * write path; without this, a stale cache line could mask the
+     * fresh data. Same pattern as ga10b_submit_and_poll's poll
+     * invalidate. */
+    const void *src = (const void *)(uintptr_t)last->output_phys;
+    if (gsp_platform && gsp_platform->cache_invalidate) {
+        gsp_platform->cache_invalidate((void *)src, cap);
+    }
+    memcpy(out, src, cap);
+    return (int)cap;
+}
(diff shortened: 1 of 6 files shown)


johnjezl:
M7 commits landed on main as 895119b via PR #376 (rebase merge). This PR's HEAD ref (eef7617) was the pre-rebase SHA — content is identical, the PR system just doesn't auto-detect rebased landings. Closing rather than merging to avoid duplicating commits.

Issue #371, opened 2026-04-25
dynamic-kernel-replace Stage 5: hardware validation on pi-5-1


johnjezl:
Hardware validation of the dynamic kernel replace stack on `pi-5-1`. Maps to **Stage 5** in `docs/dynamic-kernel-replace-plan.md`. Blocked on hardware availability and on Stages 1–4 landing first.

## Scope (sub-tasks)

1. **Sparse-file Makefile band-aid (FIRST, no hardware)** — Scope A of #392. Replace the silent `truncate -s 4G` with a sanity check that emits a clear error pointing at #392 if the build filesystem can't take the image. Pure ergonomics — converts a confusing ENOSPC into a documented failure mode. **Scope B (auto-fallback to a smaller image) is the target solution; Scope A here is a band-aid only.** #392 stays open until Scope B lands.
2. **Pi 5 SDHCI smoke test** — boot a build of `main` post-Stage-3 (#369 / PR #389) on `pi-5-1` and call `sdhci_create_bcm2712()` from a temporary debug shell command. Confirm: (a) `sdhci_create()` doesn't return NULL, (b) the probe log shows non-zero `card capacity` and the `card type =` line for the actual lab card. Surfaces any BCM2712 cfginit / CPRMAN clock-gate dependency before the diagnostic-kernel work below — if `pcie1` enumeration regressed from PR #389's PCIe core changes, this is where it shows up. ~1 hour wall time.
3. **Diagnostic kernel** — dump, at EL2 entry: EMMC2 controller clock-gate state, reset-status, controller mode, and CMD0 response. Records what VC firmware leaves behind after handoff. Resolves the empirical side of Risk 1.
4. **Tryboot round-trip** — verify the Stage 1 mailbox tags cause firmware to honor the `[tryboot]` section on the next boot. Run `labctl boot_test --count 10` to catch flakes.
5. **Real-card SDHCI quirks** — apply the BCM2712 cfginit / clock-gate / `clk_get` details that QEMU doesn't expose. Read / write / verify loops on a guarded region of the boot partition.
6. **Full round-trip via labctl** — `kernel stage` → `kernel activate` → `kernel promote` → `kernel rollback`. Closes #35 when this passes.
[...]

Issue #370, opened 2026-04-25
dynamic-kernel-replace Stage 4: kernel admin command surface


johnjezl:
Build the admin command surface for staged kernel replacement. Maps to **Stage 4** in `docs/dynamic-kernel-replace-plan.md`.

## Scope

Shell commands, admin-gated (same gate as the telnet admin path):

- `kernel status` — show current staged / active / previous image state, derived from which files exist on the FAT partition.
- `kernel stage <vfs-path>` — copy a VFS file to the FAT partition as `tryboot.img`; write a SHA-256 sidecar.
- `kernel activate` — call `bcm_mailbox_set_reboot_flags(1)` + `bcm_mailbox_notify_reboot()` (Stage 1), then `psci_system_reset()`.
- `kernel promote` — rename `tryboot.img` over `kernel_2712.img`.
- `kernel rollback` — delete `tryboot.img`; clear the flag if still set.

State machine: `empty → staged → armed → promoted` (or `rolled_back`). State inferred from FAT file existence — no separate persistence layer.

## Tests

- QEMU end-to-end: synthesize a candidate kernel blob, run `stage` → `activate` → `promote` → `rollback` against the QEMU SD card, verify FAT state transitions at each step.
- Negative cases: `stage` without source file, `activate` without a staged file, `promote` without armed state, `rollback` after `promote`.

## Dependencies

- Stage 1 (mailbox tag helpers).
- Stage 2 (FatFs).
- Stage 3 (SDHCI driver replaces FatFs stub).

## Note

Until Phase F2 (native upload) of `docs/dynamic-policy-model-loading-plan.md` lands, `kernel stage` reads from a stubbed VFS source (e.g., a baked-in test blob).

## Reference

`docs/dynamic-kernel-replace-plan.md` §3. Plan estimate: 2–3 d.

johnjezl:
Stage 4 merged via PR #395 (`3f2d6ea`). `kernel` shell admin command (status / stage / activate / promote / rollback) with streaming SHA-256 + tryboot.sha sidecar, mailbox-driven boot-flag set, PSCI reset on Pi 5. Backed by `test_suite_kernel_cmd` (10/10) plus `test_suite_sha256` (5/5 NIST FIPS 180-4 vectors). Belt-and-suspenders coverage on current main.

PR #393, opened 2026-04-25
Makefile: clearer error on SDHCI test-image creation failure (Scope A of #392)

### LINKED ISSUE #392: SDHCI test image: 4 GB sparse file requirement on the build filesystem
The QEMU SDHCI test path in `Makefile` (added by PR #389 for the dynamic-kernel-replace Stage 3 driver, #369) creates a 4 GB sparse file at `build/kernel-test/sdhci-test.img` via `truncate`. This works invisibly on every standard Linux/macOS build environment but has subtle failure modes on a few edge-case filesystems. Filing this so the constraint is discoverable when someone hits it.

## Background

- `make test` on `PLATFORM=QEMU_VIRT` runs `truncate -s 4G build/kernel-test/sdhci-test.img` if the file doesn't exist.
- QEMU is invoked with `-device sdhci-pci -drive id=...,format=raw,file=...,if=none -device sd-card,drive=...`. The card image is the backing file.
- The SDHCI test suite's `test_fatfs_round_trip_on_sdhci` formats the card with FAT32 + MBR and writes a tryboot.img + kernel_2712.img round-trip. Other SDHCI tests just write/read at single blocks.

## Why 4 GB?

QEMU's `sd-card` device returns the CCS (Card Capacity Status) bit cleared in ACMD41 — i.e. emulates SDSC, byte-addressed — for backing files below a size threshold. Our SDHCI driver supports both SDSC and SDHC since the 2nd review of #389, but the FatFs end-to-end test was authored against SDHC (block-addressed) and is the trodden path. **4 GB safely clears QEMU's CCS threshold and keeps the test on the SDHC code path.** Lowering to e.g. 1 GB still works for the *driver* but pushes the FatFs test onto a less-exercised path.

## What "sparse" means here

`truncate -s 4G` records "this file is 4 GB long" in the inode but allocates no data blocks. Reads of unallocated regions return zeros; writes allocate blocks on demand. So:

- Logical size: `stat | ls -la` → 4 GB
- Physical size after a typical `make test` run: ~500 KB to ~5 MB (only the FAT metadata + test payloads get allocated)
- Compare with `dd if=/dev/zero bs=1M count=4096` which would actually consume 4 GB.

## Filesystem matrix

| Filesystem | Sparse support | Outcome on `truncate -s 4G` | Test outcome |
|---|---|---|---|
| ext4 | ✅ | Logical 4 GB, ~0 B physical | All pass |
| btrfs | ✅ | same | All pass |
| xfs | ✅ | same | All pass |
| zfs | ✅ | same | All pass |
| APFS (macOS) | ✅ | same | All pass |
| NTFS via WSL2 | ✅ | same | All pass |
| tmpfs | ⚠️ Conditional | Logical 4 GB if size cap allows, no RAM cost until written; subsequent writes count against the cap | All 6 SDHCI tests pass IF cap ≥ 5 MB; `test_fatfs_round_trip_on_sdhci` fails first if cap < ~5 MB |
| FAT32 / exFAT | ❌ | Tries to allocate 4 GB of real bytes; ENOSPC on filesystems with < 4 GB free | Build fails before any test runs |

## Failure modes (will tests fail without sparse?)

Yes — three distinct failure paths:

1. **`truncate` refuses up front** — no sparse support AND not enough disk for 4 GB real bytes (e.g. FAT32 build dir with < 4 GB free, or a strict `ulimit -f`). Build fails at the `truncate` step. **All 6 SDHCI tests skip with a build failure.**

2. **Backing-file writes hit a quota mid-test** — possible on tmpfs with a small cap. Trivial tests need only ~16 KB of allocated bytes; `test_fatfs_round_trip_on_sdhci` needs ~1–5 MB for the partition table + FAT copies + payloads. Failure mode: `FR_DISK_ERR` from FatFs's `f_mkfs` or `f_write`. **Only the FatFs test fails; the other 5 SDHCI tests pass.**

3. **Image accumulates physical bytes across runs** — each `make test` writes more allocated blocks. Without `make kernel-test-clean` the on-disk footprint grows toward the 4 GB logical size. A long-running CI without periodic cleanup eventually behaves like failure mode #1.

## Mitigations

- **Standard dev environments (ext4 / btrfs / xfs / zfs / NTFS):** no action needed.
- **tmpfs build dir with constrained cap:** override `SDHCI_TEST_IMG_SIZE` in the Makefile to fit the cap (Makefile comment notes this). Going below ~1 GB drops QEMU into SDSC mode — most tests still pass via the SDSC code path, but the FatFs round-trip path becomes less verified.
- **FAT32 / exFAT build dir:** unsupported. Either move the build dir to a sparse-aware filesystem or set `SDHCI_TEST_IMG_SIZE` to whatever the FS can fit (with the same SDSC caveat as above).
- **CI:** ensure `make kernel-test-clean` runs between aggregations, or set `SDHCI_TEST_IMG` to a host-tmpfs path that gets wiped between jobs.

## References

- Driver: `kernel/drivers/sdhci.c` (PR #389, Stage 3 of dynamic-kernel-replace).
- Tests: `kernel/tests/test_sdhci.c`.
- Image rule: `Makefile:739-789` (search for `SDHCI_TEST_IMG`).
- The CCS-based SDHC/SDSC selection added in the 2nd review round of #389 means the **driver itself** works on either side of the threshold; this issue is only about the **test image** size choice.

## Why P3-low

Only affects exotic build environments. Standard SLM-OS dev hosts (the existing capstone Linux machines) all use ext4 / btrfs and are unaffected. There's a documented workaround (lower the image size) for tmpfs cases. No production hardware path is involved — this is purely about the QEMU test fixture.

johnjezl:
First sub-task of Stage 5 (#371) per the dynamic-kernel-replace plan. **Scope A** of #392 — band-aid only.

## Summary

Wraps the silent `truncate -s 4G` for the SDHCI test image with a clear error path. Unchanged on the happy path; convertes a confusing ENOSPC into a documented failure mode that points at #392.

## What it does

- `truncate` writes to `$@.tmp` first; on success, atomically renames to the final path. On failure, removes `.tmp` and prints a diagnostic.
- Probes the build-dir filesystem via `stat -f -c %T`:
  - **vfat / msdos / exfat** → "FAT-family filesystems don't support sparse files"
  - **tmpfs** → "likely hit its size cap; lower SDHCI_TEST_IMG_SIZE to ≥ 2 GB"
  - **other** → "disk likely doesn't have 4 GB free, or a file-size ulimit is restricting truncate"
- Always cross-references #392 for the full failure-mode matrix.

## Why this is a band-aid

**Scope B is the target solution.** The driver itself already supports SDSC mode (added in the 2nd review of PR #389), so an auto-fallback to a 1 GB image when sparse is unavailable is structurally feasible without driver changes. That's the work that closes #392.

This PR converts a confusing failure into a clear one, nothing more. #392 stays open until Scope B lands.

## Test plan

- [x] `make test` happy path unchanged: 6/6 SDHCI + 11/11 FAT32 + 7/7 mailbox + 5/5 PCIe pass
- [x] Pre-existing failures (2 eviction, 3 USB) unchanged
- [ ] Manually verified by inspection: shell logic is straight-line, `.tmp` rename is atomic, fstype probe falls back to "unknown" if `stat -f` fails. Runtime verification of the failure path requires an FAT32 build dir or a ulimit-restricted shell — not standard in our dev environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -781,7 +781,30 @@ endif
 $(SDHCI_TEST_IMG): | $(KERNEL_TEST_BUILD_DIR)
 	@if [ ! -f $@ ]; then \
 		echo "Creating sparse $@ ($(SDHCI_TEST_IMG_SIZE), SDHC-sized)"; \
-		truncate -s $(SDHCI_TEST_IMG_SIZE) $@; \
+		if ! truncate -s $(SDHCI_TEST_IMG_SIZE) $@.tmp 2>/dev/null; then \
+			rm -f $@.tmp; \
+			fstype=$$(stat -f -c %T $$(dirname $@) 2>/dev/null || echo unknown); \
+			echo ""; \
+			echo "ERROR: cannot create $(SDHCI_TEST_IMG_SIZE) sparse image at $@"; \
+			echo "       build-dir filesystem: $$fstype"; \
+			case "$$fstype" in \
+				vfat|msdos|exfat) \
+					echo "       FAT-family filesystems don't support sparse files;" ;\
+					echo "       truncate would have to allocate the full size for real." ;; \
+				tmpfs) \
+					echo "       tmpfs likely hit its size cap on the truncate write."; \
+					echo "       Lower SDHCI_TEST_IMG_SIZE in the Makefile (≥ 2 GB to" ;\
+					echo "       keep QEMU's sd-card model in SDHC mode)." ;; \
+				*) \
+					echo "       Disk likely doesn't have $(SDHCI_TEST_IMG_SIZE) free, or a" ;\
+					echo "       file-size ulimit is restricting truncate." ;; \
+			esac; \
+			echo "       See https://github.com/SLM-OS/SLM-Operating-System/issues/392"; \
+			echo "       for the full failure-mode matrix and workarounds."; \
+			echo ""; \
+			exit 1; \
+		fi; \
+		mv $@.tmp $@; \
 	fi
 
 $(KERNEL_TEST_BUILD_DIR):


johnjezl:
Superseded — combining Scope A of #392 with Stage 4 (#370) into a single PR per project direction. Branch preserved (per never-delete-branches preference); commit `cc00d59` cherry-picks cleanly into the Stage 4 branch.

The Scope A diff is unchanged in the combined PR; only the surrounding Stage 4 code gets added on top.

PR #410, opened 2026-04-26
shell: group + alphabetize help output, fix kprintf %llu

johnjezl:
## Summary

Two related fixes packaged together because both touch the shell command surface and were caught in the same investigation.

**1. `help` is no longer a wall of text.** Previously printed every command in registration order, ~50 entries with no structure. Now groups by category (Shell / Filesystem / System info / Processes & scheduling / Components & messaging / Scripting & programs / Network / Hardware control & diagnostics), alphabetized within each group. Source-side `builtin_commands[]` follows the same convention; a regression test enforces that the source layout doesn't drift away from the help output.

**2. `kprintf` now actually formats `%llu`.** `netstat` and 11 other `%llu` / `%lld` / `%llx` callers across 7 files were printing the format spec literally because the length-modifier loop only consumed a single `l`. One-line kprintf fix unblocks all of them; the alternative was changing every caller's format string.

Closes the netstat-prints-`%lu`-literally bug noticed by John during this iteration.

### `help` output

**Before** (registration order, 50+ entries, no structure):
```
Available commands:

  help       List available commands
  mem        Show memory statistics
  tasks      List all tasks
  cpu        Show CPU status
  uptime     Show system uptime
  vmm        Show virtual memory info
  ipc        Show IPC statistics
  model      Model management ...
  dtb        Show device tree info
  gpu        Show GPU info ...
  peek       Read physical memory ...
  poke       Write 32-bit word ...
  elftest    Test ELF loader
  ... (jumbled order through ~50 entries)
```

**After** (grouped by category, alphabetized within):
```
Available commands:

Shell:
  clear        Clear screen
  help         List available commands
  reboot       Restart the system

Filesystem:
  append       Append to file (append <path> <content>)
  cat          Show file contents (cat <path>)
  cd           Change directory (cd [path])
  cp           Copy file (cp <src> <dst>)
  ...

System info:
  cpu          Show CPU status
  dtb          Show device tree info
  ipc          Show IPC statistics
  mem          Show memory statistics
  ...

(Processes & scheduling, Components & messaging, Scripting & programs,
 Network, Hardware control & diagnostics — all sections similar)

Use 'help <cmd>' for detailed help on a command.
```

### `netstat`

**Before** (from John's report):
```
slmos> netstat
Network Statistics:
  RX packets: %lu  bytes: %lu
  TX packets: %lu  bytes: %lu
  RX errors:  %lu  dropped: %lu  no_buffers: %lu
  TX errors:  %lu
```

**After**:
```
slmos> netstat
Network Statistics:
  RX packets: 0  bytes: 0
  TX packets: 0  bytes: 0
  RX errors:  0  dropped: 0  no_buffers: 0
  TX errors:  0
```

## Files

| File | Purpose |
|---|---|
| `kernel/include/shell.h` | New `shell_cmd_category_t` enum + `category` field on `shell_cmd_t` |
| `kernel/src/shell.c` | `builtin_commands[]` reorganized; convention comment block enforced by regression test |
| `kernel/src/shell_sys.c` | `cmd_help` rewritten to group + alphabetize at runtime |
| `kernel/src/lua_shell.c`, `net_shell.c`, `kernel_cmd.c`, `arch/x86_64/{nvidia_gpu,pci}.c`, `ai_accel/hailo/hailo_shell.c` | External registrations now set `.category` |
| `kernel/src/kprintf.c` | Length-modifier loop accepts `ll` and `z` |
| `kernel/tests/test_shell.c` | 4 new tests (see below) |
| `kernel/tests/test_shell_session.c` | Test fixtures get explicit `.category` for completeness |
| `docs/shell.md` | Updated `help` sample output (twice), added "Adding a new command" section, refreshed `Implementation` code sample |
| `docs/api/kernel.md` | `uart_printf` format-spec line lists `%ll` and `%z` |

## New tests

| Test | What it pins |
|---|---|
| `test_kprintf_long_long` | `%llu` / `%lld` / `%llx` round-trip + the literal failing string from the netstat report |
| `test_kprintf_long_long_edge_cases` | Width modifiers (`%12llu`, `%016llx`), multiple `%llu` in one format, mixed `%d`/`%llu`/`%s` |
| `test_kprintf_size_t` | `%zu` with both small and >32-bit values |
| `test_builtin_commands_grouped_and_sorted` | Walks `builtin_commands[]` and asserts categories are contiguous, names alphabetized within each category, all categories in range. **This is the source-invariant test that enforces the convention** — drift fails the test at next `make test`. |

## Test plan

- [x] `make test` (QEMU_VIRT ARM64): exit 2, **1879 [PASS]**, 4 [FAIL]. The 4 failures are pre-existing on `origin/main` (eviction blob-validation regressions: `test_blob_validate_rejects_invalid_outer_header_fields`, `test_blob_stage_rejects_invalid_payload_headers` — neither related to this change). `origin/main` baseline is 1875 [PASS], 4 [FAIL]. **Delta: +4 PASS, +0 FAIL.**
- [x] `make kernel PLATFORM=RASPI5` — clean.
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO` — clean.
- [x] All four new tests pass.
- [x] Ran `cmd_help` in QEMU during the test boot — verified grouped output matches the documented sample.

## Pre-existing failures observed (unrelated)

```
[FAIL] test_blob_validate_rejects_invalid_outer_header_fields
[FAIL] test_blob_stage_rejects_invalid_payload_headers
```

Both fire from `kernel/tests/test_eviction.c` and reproduce identically on `origin/main` HEAD. Worth a separate follow-up issue.

### DIFF
--- a/docs/api/kernel.md
+++ b/docs/api/kernel.md
@@ -657,7 +657,7 @@ Send a null-terminated string. Synchronized. Converts `\n` to `\r\n`.
 ```c
 int uart_printf(const char *fmt, ...);
 ```
-Formatted output (synchronized). Supports `%c`, `%s`, `%d`, `%u`, `%x`, `%X`, `%p`, `%l` modifier, `%%`.
+Formatted output (synchronized). Supports `%c`, `%s`, `%d`/`%i`, `%u`, `%x`/`%X`, `%p`, `%%`. Length modifiers: `%l` and `%ll` (both treated as 64-bit on the kernel targets — `long` and `long long` are 64-bit on aarch64-elf and x86_64-elf), and `%z` for `size_t`. Width and zero-pad flags are supported (e.g. `%016llx`, `%-12s`).
 
 ```c
 int uart_snprintf(char *buf, size_t size, const char *fmt, ...);
--- a/docs/shell.md
+++ b/docs/shell.md
@@ -25,51 +25,116 @@ The industrial IoT deployment doesn't involve humans typing commands, so this is
 
 ## Commands
 
+`help` (no argument) groups commands by category and alphabetizes
+within each category. Empty categories (e.g. `Network` on a build
+without `ENABLE_NETWORKING`) are skipped. Categories are defined by
+`shell_cmd_category_t` in `kernel/include/shell.h`; the order they
+appear here matches the enum order. The source-side
+`builtin_commands[]` array is also kept grouped + alphabetized so
+"where is command X registered" and "where does X show up in `help`"
+give the same answer; the regression test
+`test_builtin_commands_grouped_and_sorted` enforces that invariant.
+
 ```
 SLM-OS> help
 Available commands:
-  help      - List available commands
-  mem       - Show memory statistics
-  tasks     - List all tasks
-  cpu       - Show CPU status
-  uptime    - Show system uptime
-  vmm       - Show virtual memory info
-  ipc       - Show IPC statistics
-  model     - Model management (load/unload/info/list/gpu/pools/stats/bench)
-  dtb       - Show device tree info
-  elftest   - Test ELF loader
-  run       - Run a program (run <name>)
-  kill      - Terminate a task by ID
-  ls        - List directory (ls [path])
-  cd        - Change directory (cd [path])
-  pwd       - Print working directory
-  cat       - Show file contents (cat <path> [offset] [length])
-  write     - Write to file (write <path> <content>)
-  mkdir     - Create directory (mkdir <path>)
-  rm        - Remove file/dir (rm <path>)
-  mv        - Move/rename (mv <src> <dst>)
-  df        - Filesystem stats (df [path])
-  truncate  - Truncate file (truncate <path> <size>)
-  append    - Append to file (append <path> <content>)
-  cp        - Copy file (cp <src> <dst>)
-  touch     - Create empty file (touch <path>)
-  stat      - Show file info (stat <path>)
-  tree      - Recursive directory listing (tree [path])
-  wc        - Count lines/words/bytes (wc <path>)
-  hexdump   - Hex dump file (hexdump <path> [offset] [len])
-  grep      - Search in file (grep <pattern> <path>)
-  find      - Find files (find <path> <pattern>)
-  component - Component system (list/run/swap/register/status)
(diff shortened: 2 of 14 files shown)


johnjezl:
## Round-1 fixes (commit `d3e9593`)

All 5 actionable items from the round-1 review addressed.

| # | Sev | Item | Fix |
|---|-----|------|-----|
| 1 | Warning | `HELP_GATHER_MAX` silent truncation | Decoupled the bounds check from the loop condition; surplus entries set a `truncated` flag that prints `"[warning: this category exceeded HELP_GATHER_MAX=N — increase the cap]"` under the affected category. |
| 2 | Warning | `category` partial-init foot-gun | Explicit "REQUIRED" comment over the field in `shell.h` calling out the silent-default-to-`SHELL_CAT_SHELL` behaviour. |
| 3 | Suggestion | `is_long` naming | Renamed to `arg_is_64bit` in `kprintf.c`. The flag now controls whether to extract a 64-bit `va_arg` from any of three paths (`%l`, `%ll`, `%z`). |
| 4 | Suggestion | drop unnecessary `(int)` cast | **Tried — caused a build break.** The cast was load-bearing: under `-Werror=sign-compare` GCC treats the unscoped enum as unsigned and rejects the `enum vs. int cat` comparison without a conversion. **Restored the cast and added an inline comment explaining why** so the next reviewer doesn't repeat the suggestion. |
| 5 | Suggestion | extend convention test to externals | New `test_external_commands_categories_in_range` walks every `shell_register_command()` entry and asserts the category is `< SHELL_CAT_COUNT`. Catches the partial-init foot-gun (item 2) on the runtime-registered side, where source-array provenance is lost so the full grouped-and-sorted rule isn't observable. |

### Verification

- `make test` on QEMU_VIRT: exit 2, **1880 [PASS]** (up from 1879 — new test #5), 4 [FAIL] (same pre-existing eviction failures from origin/main).
- All five new tests in this PR now pass: `test_builtin_commands_grouped_and_sorted`, `test_external_commands_categories_in_range`, `test_kprintf_long_long`, `test_kprintf_long_long_edge_cases`, `test_kprintf_size_t`.

PR is at 2 commits, mergeable, **not merged** — waiting on go-ahead.

PR #429, opened 2026-04-26
help: add 28 missing entries + audit existing + regression test

johnjezl:
## Summary

The shell's `help <cmd>` system was broken for many commands — anything added since the original `help.c` was written had no entry, so `help <cmd>` printed "No help available for '<cmd>'" instead of usage text.

## Repro (before)

```
slmos> help bench
No help available for 'bench'
Command returned error: -1
slmos> help admin
No help available for 'admin'
slmos> help imx219
No help available for 'imx219'
slmos> help telemetry
No help available for 'telemetry'
```

`help` (no argument) worked fine — that path iterates the live `builtin_commands[]`/`external_commands[]` tables and was unaffected. The breakage was exclusively in the per-command help text store.

## After

```
slmos> help bench
bench - Performance benchmarks

Usage:
  bench <subcommand> [args...]

Subcommands:
  context        Context-switch latency (create/switch/destroy)
  irq            Interrupt latency (timer tick interval accuracy)
  ipc            IPC message round-trip latency
  ...

slmos> help admin
admin - Seven-page operator TUI for live observability
...
```

## Audit

Walked the live registration tables and diff'd against `help_entries[]`:

- **28 commands missing entries:** `admin`, `bench`, `bpmp`, `diag`, `eviction`, `gpu`, `hailo`, `hspdiag`, `imx219`, `kernel`, `lua-admin`, `macbdiag`, `msg`, `nvgpu`, `pci`, `pcietrain`, `peek`, `poke`, `rtldiag`, `sched`, `sleep`, `tcpsh`, `telemetry`, `telnetd`, `timdiag`, `top`, `xhci`, `xhcidiag`.
- **0 orphaned entries** (no help texts for commands that no longer exist).
- **1 stale entry:** `component` listed only `register`/`unregister`/`status`/`list` but the actual command also accepts `builtins`/`run`/`swap`/`send`. Updated.
- **38 existing entries audited and current:** spot-checked against live command behaviour.
- **1 follow-up worth tracking:** the `lua` entry's massive `slm.*` binding list has accumulated drift since it was last touched. Out of scope for this fix; a rewrite that pulls the list from the live registry would be a cleaner solution. The entry is still functional, just not exhaustive.

## Files

| File | Purpose |
|---|---|
| `kernel/src/help.c` | +29 HELP_TEXT entries (28 new + `component` rewrite) with section-marker comment pointing at the regression test |
| `kernel/tests/test_shell.c` | New `test_every_command_has_help_entry` regression test — walks live tables and asserts every command has an entry. `#include "../include/help.h"` added. |

## New regression test

`test_every_command_has_help_entry` walks both `builtin_commands[]` and `external_commands[]` at runtime and calls `help_exists(name)` for each. Test fixtures (`testcmd`, `t_nested_inner`, `t_nested_outer`) registered transiently by other tests are skipped — the convention only applies to production registrations. The test surfaces a short `[FAIL] no help entry for <cmd>` line per missing entry on the harness UART, so future drift is debuggable from the boot log alone (rather than a single failure that only names the first miss).

The test caught one I missed in my initial audit (`kernel`, registered by `kernel_cmd.c` for the dynamic-replace tryboot workflow) — exactly the failure mode it's designed to prevent.

## Test plan

- [x] `make test` (QEMU_VIRT ARM64): exit 2, **1953 [PASS]**, 4 [FAIL]. Up by 5 PASS from `origin/main` baseline (1948). The 4 failures are pre-existing eviction-blob-validation tests on `origin/main`, unrelated.
- [x] `test_every_command_has_help_entry`: PASS for all built-in and external commands present in this build.
- [x] Live verification: drove `help bench`, `help admin`, `help imx219`, `help telemetry`, `help peek` in QEMU — all produce real content. `help nonsense_command` still correctly says "No help available" (negative-test path).
- [x] `make kernel PLATFORM=RASPI5`: clean.
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO`: clean.


johnjezl:
## Round-1 fixes (commit `6229b0f`)

All 3 suggestions from the review addressed.

| # | Item | Fix |
|---|------|-----|
| 1 | `kernel` placed before `imx219` in the alphabetized block | Moved `HELP_TEXT("kernel", ...)` to its alphabetical position between `imx219` and `lua-admin`. The block now honours its own "kept alphabetized" claim. |
| 2 | `eviction model status\|load\|...` was vague | Enumerated the actual subcommands: `status`, `load <kind> <path>`, `activate <kind>`, `rollback <kind>`, `clear <kind>`, `autoload <status\|set\|clear>`. Source-of-truth: `cmd_eviction` in shell_sys.c. |
| 3 | Hardcoded test fixture skip-list (3 names) | Switched to the `t_` prefix convention: `if (strncmp(cmd->name, "t_", 2) == 0) continue;`. Renamed the lone non-prefixed fixture `testcmd` → `t_testcmd`; the other two (`t_nested_inner`, `t_nested_outer`) already use the prefix. Convention is documented at both the registration site and the skip site. |

### Verification

- `make test` (QEMU_VIRT ARM64): 1953 [PASS], 4 [FAIL] (same pre-existing eviction-blob-validation baseline; unchanged from prior commit).
- `test_every_command_has_help_entry`: PASS.
- `test_shell_register_external_command` (renamed-fixture path): PASS.

PR is at 2 commits, mergeable, **not merged** — waiting on go-ahead.

PR #432, opened 2026-04-26
test_eviction: fix MLP payload buffer undersize

johnjezl:
## Summary

Fixes the two long-standing failures on `make test`:

- `test_blob_validate_rejects_invalid_outer_header_fields`
- `test_blob_stage_rejects_invalid_payload_headers`

Both have been failing on `origin/main` for several PRs in a row. They share the same root cause: a `static uint8_t mlp_payload[2048]` buffer passed to `build_eviction_mlp_payload`, which actually requires **17 676 bytes** (8-byte header + 4417 fp32 weights for the 27→64→32→16→1 MLP). The helper bails with `return 0`, the very next line `TEST_ASSERT_TRUE(payload_len > 0)` fires, test fails.

## Fix

- Hoist `EVICTION_MLP_PAYLOAD_BYTES` (= 17 676) out as a named constant near the helper. Both buffer declarations now reference it.
- Add `static_assert(TOTAL == EVICTION_MLP_PAYLOAD_BYTES, ...)` inside `build_eviction_mlp_payload` so any future MLP-topology change fails the build instead of silently regressing the buffers.

The buffers are `static` (BSS), so the size bump from 2 KB to ~17 KB has no impact on the 16 KB kernel stack.

## Verification

- `make test` (QEMU_VIRT ARM64): exit 0, **1969 [PASS], 0 [FAIL]**. First all-green `make test` on `origin/main` in weeks.

## Note

`kernel/tests/test_blob_autoload.c` has an independent copy of `build_eviction_mlp_payload` with the same 2 KB buffer pattern (line 964). Its caller is gated on `CONFIG_AI_SCHEDULER` so it doesn't run on the default `make test`, and the `static_assert` added here only covers the test_eviction.c copy. Fix-up for that file would be welcome as a follow-up but out of scope here.

### DIFF
--- a/kernel/tests/test_blob_autoload.c
+++ b/kernel/tests/test_blob_autoload.c
@@ -39,11 +39,23 @@ static uint32_t fnv1a32(const uint8_t *data, size_t len)
     return hash;
 }
 
+/* Outer-blob header size written by `build_outer_blob`. Same value and
+ * intent as BLOB_TEST_OUTER_HEADER_BYTES in test_eviction.c — see the
+ * consolidation follow-up tracked in PR #432. */
+#define BLOB_OUTER_HEADER_BYTES 24
+
 static size_t build_outer_blob(uint16_t kind_id, const uint8_t *payload,
                                size_t payload_len, uint8_t *out, size_t out_cap)
 {
+    /* Tripwire: any new offset written below must bump
+     * BLOB_OUTER_HEADER_BYTES. */
+    static_assert(BLOB_OUTER_HEADER_BYTES == 24,
+                  "build_outer_blob writes 24 bytes at out[0..23]; "
+                  "BLOB_OUTER_HEADER_BYTES and the writes below "
+                  "must move together");
+
     uint32_t checksum = fnv1a32(payload, payload_len);
-    size_t total = 24 + payload_len;
+    size_t total = BLOB_OUTER_HEADER_BYTES + payload_len;
     if (out_cap < total) return 0;
 
     out[0] = 'S'; out[1] = 'E'; out[2] = 'M'; out[3] = 'B';
@@ -61,7 +73,7 @@ static size_t build_outer_blob(uint16_t kind_id, const uint8_t *payload,
     out[18] = (uint8_t)((checksum >> 16) & 0xFF);
     out[19] = (uint8_t)((checksum >> 24) & 0xFF);
     out[20] = 0; out[21] = 0; out[22] = 0; out[23] = 0;
-    memcpy(out + 24, payload, payload_len);
+    memcpy(out + BLOB_OUTER_HEADER_BYTES, payload, payload_len);
     return total;
 }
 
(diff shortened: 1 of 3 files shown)


johnjezl:
## Round 1 review fixes — pushed in 3dd1afa

Both actionable items from the review are addressed:

**1. `+ 64` magic-number in `test_eviction.c` blob buffers**

Replaced with `BLOB_TEST_OUTER_HEADER_BYTES` (= 24, matching the actual outer-blob header layout). Added a `static_assert` inside `build_test_blob` so any future change to the header layout has to update both the constant and the literal byte offsets together — drift fails the build instead of silently corrupting blobs.

**2. Sibling helper in `test_blob_autoload.c` (acknowledged follow-up)**

Pulled the same pattern into the sibling. `build_outer_blob` gets `BLOB_OUTER_HEADER_BYTES` + the static_assert tripwire. The CONFIG_AI_SCHEDULER-gated `build_eviction_mlp_payload` helper gets `EVICTION_MLP_PAYLOAD_BYTES` (= 17676 — same value the test_eviction.c sibling uses) and resizes the two callers' buffers (`ev_mlp_payload[2048]` → `[EVICTION_MLP_PAYLOAD_BYTES]`, `blob[8192]` → `[EVICTION_MLP_PAYLOAD_BYTES + BLOB_OUTER_HEADER_BYTES]`) which were silently undersized.

## Verification

- `make test` (default): **PASS 1910 / FAIL 0**, exit 0. Both target tests show `[PASS]`:
  - `test_blob_validate_rejects_invalid_outer_header_fields`
  - `test_blob_stage_rejects_invalid_payload_headers`

## Note on `make test AI_SCHED=ON`

Pre-existing unrelated issue: the AI_SCHED=ON test kernel exceeds the 16 MiB linker ceiling on `origin/main` (link fails with `Kernel image too large!`). Filed as issue #436. The CONFIG_AI_SCHEDULER-gated `test_blob_autoload.c` fix in this PR is therefore latent — it will become reachable once #436 is resolved. Not blocking this PR.

johnjezl:
## Round 2 review fixes — pushed in eb5cbc8

Both Suggestion-level items addressed:

**1. Comment overpromise on `static_assert`**

The previous comment claimed the assert "catches drift between the constant and the literal offsets." It actually only catches changing the constant without matching the writes — it does NOT catch the inverse (a maintainer adds `out[24] = ...` without bumping the constant). The comment now says so explicitly:

> Fires if a maintainer changes the constant without updating the literal byte writes below. Does NOT catch the inverse (adding a new `out[24] = ...` write without bumping the constant) — that case is on the reviewer.

**2. Cross-file constant + helper duplication**

Lifted the constants AND the helpers (`fnv1a32`, `build_outer_blob`, `build_eviction_mlp_payload`) into `kernel/tests/test_blob_helpers.h`. `static inline` in a header so each TU gets its own copy and -Werror=unused-function does not fire (which also let me drop the `#ifdef CONFIG_AI_SCHEDULER` gate around `build_eviction_mlp_payload` in test_blob_autoload.c — the gate existed solely to silence the warning).

Renames (mechanical, file-local in test_eviction.c):
- `build_test_blob` → `build_outer_blob`
- `BLOB_TEST_OUTER_HEADER_BYTES` → `BLOB_OUTER_HEADER_BYTES`

The follow-up that the round-1 commit message acknowledged ("fixing this duplication in a shared test-utility header is a tracked follow-up") is now done.

## Stats
- 3 files changed, 168 insertions, 272 deletions — **net -104 lines**
- 1 new header (143 lines), the duplication it replaces was ~250 lines

## Verification
- `make test` — PASS 1910/1910, exit 0
- Both target tests still `[PASS]`:
  - `test_blob_validate_rejects_invalid_outer_header_fields`
  - `test_blob_stage_rejects_invalid_payload_headers`

(AI_SCHED=ON test build remains broken for the unrelated reason already documented in #436.)

Issue #427, opened 2026-04-26
Jetson CDC-ECM TX path stalls after burst output (e.g. `model`); kernel stays alive on serial


johnjezl:
Reproduces on jetson-nano-2 with the current main kernel (post-#424).

**Symptom:** A telnet client connected to `192.168.4.5:2323` runs a command that emits a multi-line burst — e.g. bare `model` (~28 lines of pool/loaded-model table) — and the telnet connection appears to freeze. ping from the host returns 'Destination Host Unreachable' immediately.

**What's actually broken:** The kernel is **not crashed**. Confirmed via serial:
- `uptime` keeps incrementing
- `tasks` shows all idle/shell/net_pump tasks running normally
- `ifconfig` shows `sl0` UP, DHCP bound, IP/gw set
- BUT `ping 192.168.4.1` from inside the kernel returns 'Failed to send ping' 4× — the *outgoing* path can't transmit

From a telnet client's view this is indistinguishable from a kernel crash, but `reboot` is not needed; a power-cycle restores network and a fresh kexec is fine.

**Trigger correlation (new from this debug session):** A simple `model` command emitting ~28 lines of output reliably puts the system into this state. Smaller bursts (one-line `uptime`) do not. Suggests the TX flake is sensitive to a single tcp_write batch size or to sustained pacing.

**Reproduction steps:**
1. Boot SLM-OS via slmos-kexec on jetson-nano-2.
2. Connect via `nc 192.168.4.5 2323`.
3. Run `model`.
4. Observe ~28-line table prints fine, then the next command (or `ping`) hangs / returns DHU.
5. Switch to serial: kernel is fully responsive.

**Pre-existing scope:** Reproduces during M7 admin-telemetry hardware verification yesterday; not caused by the admin-telemetry suite — the suite's only TX is single-line emit lines plus `telemetry stats` (4 lines), which doesn't trigger it. The trigger is *any* command that emits a 20+ line burst.

**Suspect components:**
- `kernel/drivers/usb/xhci/` — XHCI bulk-OUT TRB completion handling (Tegra234 XHCI is a known #266-family flake source)
- `kernel/usb/class/cdc_ecm.c` — TX descriptor recycling
- lwIP `tcp_output` interaction with the CDC-ECM transmit ring

**Workaround:** Use serial console for sessions involving multi-line shell output. Use telnet only for short queries.

Filing as P2 because:
- Pre-existing (not regression from M1-M7 admin-telemetry work)
[...]

Issue #412, opened 2026-04-26
test_blob_validate/stage_rejects_invalid_*_headers fail on main (payload_len > 0)


johnjezl:
Two QEMU eviction tests fail on a clean main checkout (`make test` only, no special flags):

```
[FAIL] test_blob_validate_rejects_invalid_outer_header_fields
   FAILED at kernel/tests/test_eviction.c:1014 - payload_len > 0
[FAIL] test_blob_stage_rejects_invalid_payload_headers
   FAILED at kernel/tests/test_eviction.c:973 - payload_len > 0
```

Reproduces on `origin/main` HEAD (e7a7d88) with no other modifications. Likely the test fixture is constructing a blob whose `payload_len` ends up zero, which fails an early-precondition check before the validation path even runs.

Surfaced while running `make test` as part of admin & telemetry suite M1 PR. Filing so the M1 PR's CI failure is recognised as pre-existing rather than caused by the new latency_hist/rate_ewma code.

**Verification:** Re-run `make test` on a clean checkout of main; both tests fail.

PR #435, opened 2026-04-26
fix(net): protect lwIP heap with SYS_ARCH_PROTECT (#427 follow-up)

johnjezl:
## Summary

- Set `LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT=1` in `kernel/include/lwipopts.h` so the lwIP heap is actually protected.
- Root cause of the `lwIP ASSERT: invalid next ptr at .../mem.c:790` reported on jetson-nano-2 after the PR #430 pool bumps.

## Problem

`LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT` defaults to 0. With that default, `LWIP_MEM_FREE_PROTECT` (used in `mem_malloc`, `mem_free`, `mem_trim`) resolves to `sys_mutex_lock(&mem_mutex)`. Under `NO_SYS=1` that macro expands to an **empty no-op** in `lwip/sys.h`. The lwIP heap was running with zero serialisation, while shell-task lwIP entries (`cmd_ping` → `raw_sendto`, `cmd_telnetd` → `tcp_listen` / `tcp_close`) raced against the `net_pump` task's `tcp_input` / `pbuf_alloc` / timer-wheel work.

The race surfaced after PR #430 bumped `PBUF_POOL_SIZE` / `MEMP_NUM_TCP_PCB` / `MEMP_NUM_TCP_SEG` — bigger pools meant more heap traffic which widened the race window.

## Fix

Set `LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT=1`. That switches `LWIP_MEM_FREE_PROTECT` to `SYS_ARCH_PROTECT`, which already routes to `sys_arch_protect()` → `irq_save()` in `kernel/net/sys_arch.c`. The result is task-vs-task and task-vs-IRQ serialisation on the local CPU, which is sufficient because every lwIP-touching task (`net_pump`, shell, shell-tcp sessions) is pinned to CPU 0 (`cpu_affinity = 0` in `kernel/src/main.c` and `task_set_affinity(t, 0)` in `kernel/src/tcp_shell_server.c`).

## Test plan

- [x] `make kernel` (QEMU ARM64) — clean build
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON -DNET_TELNETD_AUTOSTART=ON -DNET_DHCP_AT_BOOT=ON"` — clean build
- [x] `make test` — only pre-existing #412 blob test failures
- [x] Deployed to jetson-nano-2 via `slmos-kexec`
- [x] 12 sequential `slm-put.py` 4 KB cycles — no assert
- [x] 10 parallel 64 KB `slm-put.py` cycles — no assert, all uploads verified
- [x] Mixed shell command typing during upload load — no panic
- [x] Shell-side `ping 192.168.4.1` → 4/4 replies (the path that previously raced against `net_pump`)
- [x] Host-side `ping 192.168.4.5` → 5/6 replies, kernel responsive

### DIFF
--- a/kernel/include/arch/cc.h
+++ b/kernel/include/arch/cc.h
@@ -115,8 +115,13 @@ uint32_t lwip_rand_slm(void);
 /* System Protection (NO_SYS mode)                                             */
 /* -------------------------------------------------------------------------- */
 
-/* Protection type for critical sections - not used in NO_SYS=1 mode
- * but must be defined for lwIP sys.h */
-typedef int sys_prot_t;
+/* Protection token for lwIP critical sections.
+ *
+ * `sys_arch_protect()` returns `irq_save()`'s value; on AArch64 that
+ * is the full 64-bit DAIF and on x86-64 it is the full 64-bit RFLAGS.
+ * Match `irq_flags_t` (uint64_t) so the round-trip through
+ * `sys_arch_unprotect()` carries every bit verbatim — no narrowing
+ * cast that future register extensions could quietly truncate. */
+typedef uint64_t sys_prot_t;
 
 #endif /* ARCH_CC_H */
--- a/kernel/include/lwipopts.h
+++ b/kernel/include/lwipopts.h
@@ -45,6 +45,21 @@
 /* Memory alignment (8-byte for AArch64) */
 #define MEM_ALIGNMENT               8
 
+/* Route the lwIP heap protection through SYS_ARCH_PROTECT (irq_save in
+ * sys_arch.c) instead of the sys_mutex_t path. With NO_SYS=1 the
+ * sys_mutex_* macros expand to empty no-ops (lwip/sys.h), so the default
+ * setting of 0 leaves `mem_malloc/mem_free/mem_trim` completely
+ * unprotected — every shell-task lwIP call (cmd_ping → raw_sendto,
+ * cmd_telnetd → tcp_listen/tcp_close) races against the net_pump task
+ * that drives `tcp_input`, `pbuf_alloc`, and the timer wheel. Setting
+ * this to 1 makes lwIP wrap each heap critical section in
+ * SYS_ARCH_PROTECT → sys_arch_protect() → irq_save(), which serialises
+ * task context against task context AND task against IRQ on the local
+ * CPU. Surfaced as `lwIP ASSERT: invalid next ptr at .../mem.c:790`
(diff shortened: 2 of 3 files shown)


johnjezl:
Addressed actionable items from /slmos-review (commit 5960b7f):

- **`sys_prot_t` narrowing** → widened to `uint64_t` matching `irq_flags_t`, casts dropped from `sys_arch_protect`/`sys_arch_unprotect`. Added `_Static_assert(sizeof(sys_prot_t) == sizeof(irq_flags_t))` to prevent silent re-introduction.
- **Implicit CPU 0 invariant** → documented in `kernel/net/sys_arch.c` near `sys_arch_protect()`, lists the four task creation sites it depends on, and explains what to do if a future caller runs on a different CPU.
- **No regression test** → added `_Static_assert(LWIP_ALLOW_MEM_FREE_FROM_OTHER_CONTEXT == 1)` in `sys_arch.c` so any revert of the lwipopts.h flag fails the build with a pointed message. Compile-time guard, not a runtime fuzz, but it's the deterministic regression check that's tractable here.
- **Cross-platform build verification** → all four platforms now built clean:
  - QEMU ARM64 (default) — `make kernel`
  - Jetson Orin Nano — `make kernel PLATFORM=JETSON_ORIN_NANO EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON -DNET_TELNETD_AUTOSTART=ON -DNET_DHCP_AT_BOOT=ON"`
  - Raspberry Pi 5 — `make kernel PLATFORM=RASPI5 EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON"`
  - x86-64 — `make kernel PLATFORM=X86_64 EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON"`

`make test` still passes (only pre-existing #412 blob failures, unrelated). Re-deployed the new kernel to jetson-nano-2 — 4/4 kernel-side ping replies, 5×64 KB slm-put cycles clean.

The IRQ-disable-latency-across-`mem_malloc` suggestion was left as a note rather than a code change — agree it's worth measuring under load if scheduler jitter ever spikes, but no observable issue today and there's no good knob to tune without inverting the protection model.

PR #440, opened 2026-04-27
feat(net): bump lwIP MEM_SIZE 32→128 KB and add shell-tcp leak detector

### LINKED ISSUE #442: Shell-tcp session teardown leaks heap (~1-2 KB per session)
## Summary

Per-session lwIP heap leak detector (PR #440) caught real heap leaks during shell-tcp session teardown. With `NET_SHELL_TCP_LEAK_THRESHOLD_BYTES = 1024`:

```
Shell-tcp sessions:
  opened:     10  closed: 9  active: 1  peak: 1
  heap delta: last=-1136 max=1984
  leaks:      warnings=2  total=3880 bytes
```

8 telnet open/close cycles → 2 closes exceeded the threshold (max single 1984 B). Average leak across cycles ~480 B/session, spikes to ~2 KB.

## Why it matters

The original RX-stall watchdog firing pattern:
```
[WARN] net: RX stalled (link up, 10010 ms idle, threshold 10000 ms)
[WARN] net:   rx_packets=719 dropped=66 no_buffers=0
[WARN] net:   pbuf_pool=0/64 tcp_pcb=2/32 heap=32768/32768
```

PR #440 mitigated by bumping `MEM_SIZE` 32 KB → 128 KB, raising the floor by ~4×. The leak is still there — at ~1 KB/session sustained, the new heap fills in a few hundred cycles.

## Repro

On jetson-nano-2 (or any networking-capable build):
```python
for _ in range(20):
    s = socket.create_connection(("192.168.4.5", 2323))
    time.sleep(0.4); s.close()
    time.sleep(0.2)
```

Then `netstat` on the device — `leaks: warnings=N total=M bytes` will be non-zero.

## Likely culprits to investigate

1. `tcp_write(... TCP_WRITE_FLAG_COPY)` allocations from the shell-tcp ring drain that don't free if the peer half-closes mid-write or abrupt-closes (RST).
2. `pbuf` chains held in `tcp_pcb->refused_data` after the peer abandons before `tcp_recved` consumes everything.
3. Off-by-one refcount in the `shell_io_tcp` RX ring → pbuf hand-off to the session task — if a pbuf gets ref-bumped without a matching free during teardown.
4. The per-session `tcp_shell_ctx` struct allocation itself (allocated where? when freed? need to trace).

## Next steps

- File the investigation here.
- Use `netstat` to track per-session deltas after each isolated close path (clean close vs RST vs half-close mid-write).
- Add finer instrumentation if needed (count per-allocation-type).
- Fix and verify deltas drop to ≤ a few hundred bytes per session (TIME_WAIT footprint expected).

## See also

- PR #440 — `feat(net): bump lwIP MEM_SIZE 32→128 KB and add shell-tcp leak detector`
- PR #439 — `feat(net): add RX-stall watchdog with auto-snapshot on alarm`
- `kernel/src/shell_io_tcp.c` — the TCP backend
- `kernel/src/tcp_shell_server.c` — accept callback + session task lifecycle

johnjezl:
## Summary

Two changes prompted by the natural firing of PR #439's RX-stall watchdog under multi-session telnet load, plus the follow-up fix that resolves #442:

```
[WARN] net: RX stalled (link up, 10010 ms idle, threshold 10000 ms)
[WARN] net:   rx_packets=719 dropped=66 no_buffers=0
[WARN] net:   pbuf_pool=0/64 tcp_pcb=2/32 heap=32768/32768
```

`heap=32768/32768` — the 32 KB lwIP heap was saturated and packets were dropping (~9% loss); the watchdog correctly reported it.

## 1. `MEM_SIZE` 32 KB → 128 KB

The 32 KB ceiling was the original VirtIO-net bringup default (DHCP + ping workload only) and never revisited. Multi-session telnet now creates concurrent `tcp_write(... TCP_WRITE_FLAG_COPY)` allocations + per-frame `pbuf_alloc(PBUF_RAW, len, PBUF_RAM)` on RX — both heap-backed. 128 KB gives ~80 in-flight 1500-byte pbufs of headroom.

BSS growth is +96 KB; negligible on 4-8 GB systems.

## 2. Per-session shell-tcp leak detector

New `tcp_shell_server_get_stats()` API + `netstat` block:

```
Shell-tcp sessions:
  opened:     20  closed: 19  active: 1  peak: 1
  heap delta: last=-1688 max=856
  leaks:      warnings=0  total=0 bytes
```

Each session snapshots `lwip_stats.mem.used` at session-open; on session-close the delta is computed. Deltas exceeding `NET_SHELL_TCP_LEAK_THRESHOLD_BYTES` (1024) trigger a `WARN` and accumulate into `leak_warnings` / `total_suspicious_leak_bytes`. The watchdog snapshot also gains `heap_used_peak` from `lwip_stats.mem.max`.

## 3. Fix #442: close-time measurement timing

The first revision of this PR measured the close-time delta from the session task, immediately after `shell_io_tcp_destroy`. But `shell_io_tcp_destroy` only sets flags — the actual `tcp_close(pcb)` runs on a *later* net_pump tick inside `shell_io_tcp_poll`. Between those two events lwIP still holds the unacked TX queue (normal TIME_WAIT-bound state, freed once the peer ACKs or TIME_WAIT expires). The original measurement caught that transient state and reported it as a leak.

`4f02bd9` moves the heap snapshot onto `tcp_shell_ctx` (which outlives `shell_session` by one or two net_pump ticks) and fires the close-time measurement from `shell_io_tcp_poll` *after* `tcp_close(pcb)` has run. With that timing, deltas are clean.

## Hardware verification on jetson-nano-2

| Workload | Before fix (v1 timing) | After fix (post-tcp_close) |
|---|---|---|
| 8 telnet open/close cycles | leaks=2, total=3880 B, max=1984 | leaks=0, total=0, max=856 |
| +10 slm-put 8 KB transfers | n/a | leaks=0, max=856 (unchanged) |

19 closed sessions across two workload mixes, zero false-positive WARNs, max single-session delta is 856 bytes (~17% under the 1 KB threshold), heap stays at 12.6 KB / 128 KB peak 14.4 KB. The original watchdog-firing pattern (32 KB heap saturating) is fully resolved by the heap bump alone — the leak detector was a useful red herring that produced a real-looking signal until the measurement timing was corrected.

## Test plan

- [x] `make kernel` (QEMU ARM64) — clean
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON -DNET_TELNETD_AUTOSTART=ON -DNET_DHCP_AT_BOOT=ON"` — clean
- [x] `make kernel PLATFORM=RASPI5 EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON"` — clean
- [x] `make kernel PLATFORM=X86_64 EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON"` — clean
- [x] `make test` — only the pre-existing 19 blob-autoload failures (verified those exist on plain `origin/main` too; not introduced by this work)
- [x] Hardware verification on jetson-nano-2 (table above)

Closes #442.

### DIFF
--- a/kernel/include/lwipopts.h
+++ b/kernel/include/lwipopts.h
@@ -39,8 +39,28 @@
 #define MEMP_MEM_MALLOC             0
 #define MEMP_MEM_INIT               1  /* Zero-initialize pools at startup */
 
-/* Heap size for variable-length allocations */
-#define MEM_SIZE                    (32 * 1024)  /* 32 KB */
+/* Heap size for variable-length allocations.
+ *
+ * Bumped 32 KB → 128 KB after the RX-stall watchdog (PR #439) caught
+ * heap exhaustion under multi-session telnet + slm-put load:
+ *
+ *   [WARN] net: RX stalled (link up, 10010 ms idle, threshold 10000 ms)
+ *   [WARN] net:   rx_packets=719 dropped=66 no_buffers=0
+ *   [WARN] net:   pbuf_pool=0/64 tcp_pcb=2/32 heap=32768/32768
+ *
+ * The 32 KB ceiling was the original VirtIO-net bringup default
+ * (DHCP + ping workload only) and never revisited as the workload
+ * grew. Multi-session telnet creates concurrent
+ * `tcp_write(... TCP_WRITE_FLAG_COPY)` allocations; per-frame
+ * `pbuf_alloc(PBUF_RAW, len, PBUF_RAM)` on RX is also heap-backed.
+ * 128 KB gives ~80 in-flight 1500-byte pbufs of headroom which is
+ * comfortable for 16 simultaneous shell-tcp sessions.
+ *
+ * Tradeoffs: BSS growth is +96 KB (negligible on 4-8 GB systems);
+ * a fragmented heap walk under SYS_ARCH_PROTECT (PR #435) holds
+ * IRQs off proportionally longer. The watchdog will catch it if
+ * IRQ latency starts dropping packets at this size; revisit if so. */
+#define MEM_SIZE                    (128 * 1024)  /* 128 KB */
 
 /* Memory alignment (8-byte for AArch64) */
 #define MEM_ALIGNMENT               8
--- a/kernel/include/net.h
+++ b/kernel/include/net.h
@@ -284,6 +284,7 @@ struct net_watchdog_snapshot {
     uint16_t  tcp_pcb_used;
     uint16_t  tcp_pcb_avail;
     uint32_t  heap_used;             /* lwip_stats.mem.used */
+    uint32_t  heap_used_peak;        /* lwip_stats.mem.max — high-water mark since boot */
     uint32_t  heap_avail;            /* lwip_stats.mem.avail */
     uint32_t  stall_events;          /* count of stall→alarm transitions since boot */
     uint32_t  recovery_events;       /* count of alarm→recovery transitions since boot */
--- a/kernel/include/shell_session.h
+++ b/kernel/include/shell_session.h
@@ -99,6 +99,14 @@ struct shell_session {
      * connection so concurrent operators cannot clobber each other's
      * transfers. */
     struct shell_xput_session xput;
+
+    /* lwIP heap usage at the moment this session's task entered.
+     * Subtracted from heap_used at session-task exit to surface
+     * per-session leaks (tcp_write COPY data not freed, lingering
+     * pbufs in the unsent queue, etc). Set by tcp_shell_server's
+     * accept callback; consumed by the session task on teardown.
+     * Zero on the console session (no leak detection there). */
+    uint32_t heap_used_at_open_bytes;
 };
 
 /* Get the singleton console session (UART-backed). Always non-NULL
--- a/kernel/include/shell_session.h
+++ b/kernel/include/shell_session.h
@@ -99,14 +99,6 @@ struct shell_session {
      * connection so concurrent operators cannot clobber each other's
      * transfers. */
     struct shell_xput_session xput;
-
-    /* lwIP heap usage at the moment this session's task entered.
-     * Subtracted from heap_used at session-task exit to surface
-     * per-session leaks (tcp_write COPY data not freed, lingering
-     * pbufs in the unsent queue, etc). Set by tcp_shell_server's
-     * accept callback; consumed by the session task on teardown.
-     * Zero on the console session (no leak detection there). */
-    uint32_t heap_used_at_open_bytes;
 };
 
 /* Get the singleton console session (UART-backed). Always non-NULL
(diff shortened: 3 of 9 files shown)


johnjezl:
Added a follow-up commit (\`4f02bd9\`) that resolves #442 in the same PR.

## TL;DR

The "leak" the v1 detector reported was a measurement-timing artifact, not a real leak. With the close-time measurement moved to *after* \`tcp_close\` runs in \`shell_io_tcp_poll\`, deltas come back clean across the same workload that produced the original WARN spam.

## What was wrong

The session task called \`tcp_shell_server_note_session_close(sess)\` right after \`shell_io_tcp_destroy(io)\`, but \`shell_io_tcp_destroy\` only flips the closed/shell_done flags — the actual \`tcp_close(pcb)\` runs on a *later* net_pump tick inside \`shell_io_tcp_poll\`. Between those two events, lwIP still holds the unacked TX queue (normal TIME_WAIT-bound state). The detector saw that transient state and reported it as a leak.

## The fix

- \`heap_used_at_open_bytes\` moved from \`shell_session\` to \`tcp_shell_ctx\` (which outlives \`shell_session\` by one or two net_pump ticks). Snapshot is taken at the end of \`shell_io_tcp_create\`, after the initial telnet negotiation has been queued.
- \`tcp_shell_server_note_session_open(uint32_t session_id)\` now just bumps counters; the heap snapshot is owned downstream.
- \`tcp_shell_server_note_session_close(uint32_t session_id, int32_t heap_delta_bytes)\` accepts a precomputed delta, called from \`shell_io_tcp_poll\` immediately before \`ctx_free\`.
- The session task no longer notes the close.

## Hardware verification on jetson-nano-2

| Workload | Before (v1 timing) | After (post-tcp_close) |
|---|---|---|
| 8 telnet open/close cycles | leaks=2, total=3880 B, max=1984 | leaks=0, total=0, max=856 |
| +10 slm-put 8 KB transfers | n/a | leaks=0, max=856 (unchanged) |

19 closed sessions, zero false-positive WARNs, max single-session delta sits ~17% under the 1 KB threshold, heap stays at 12.6 KB / 128 KB (peak 14.4 KB).

The 128 KB heap bump is what actually mitigated the original watchdog-firing pattern; the leak detector was a useful red herring that produced a real-looking signal until the timing was corrected.

## Updated PR scope

| Commit | What |
|---|---|
| \`cfd2c94\` | \`MEM_SIZE\` 32→128 KB; add \`tcp_shell_server_get_stats()\` + per-session leak detector + \`heap_used_peak\` in watchdog snapshot; surface in \`netstat\` |
| \`4f02bd9\` | Move close-time measurement to post-\`tcp_close\` (the actual fix; closes #442) |

Closes #442.

Issue #533, opened 2026-04-28
12 test_blob_autoload failures regressed on main after PR #531 merge


johnjezl:
## Context

PR #531 (merge commit 30ebfaf) was reported as fixing 21 test_blob_autoload + persistent_lfs failures with \`make test\` clean on the merging environment.

Verified on a fresh QEMU_VIRT \`make test\` against the exact PR #531 merge commit (\`git checkout 30ebfaf\` → \`make test\`) on this machine: 12 test_blob_autoload failures reproduce, 0 regressions on persistent_lfs side.

Same 12 failures on \`origin/main\` HEAD (\`1073c09\` — docs-only commit on top of #531).

## Failures

\`\`\`
[FAIL] test_blob_autoload_init_marks_existing_fat_config_authoritative
[FAIL] test_blob_autoload_init_recovers_missing_fat_blob_from_lfs
[FAIL] test_blob_autoload_set_get_clear_round_trip
[FAIL] test_blob_autoload_fat_update_invalidates_lfs_fallback
[FAIL] test_blob_autoload_set_ignores_unrelated_stale_lfs_entry
[FAIL] test_blob_autoload_init_migrates_lfs_entries_to_fat
[FAIL] test_blob_autoload_set_migrates_all_lfs_entries_to_fat
[FAIL] test_blob_autoload_shell_commands
[FAIL] test_blob_autoload_overwrites_existing_conf
[FAIL] test_blob_autoload_recovers_from_backup_conf
[FAIL] test_blob_boot_autoload_skips_tampered_managed_blob
[FAIL] test_blob_autoload_rejects_malformed_metadata_suffix
\`\`\`

## Hypothesis

Either:
1. \`make test\` is non-deterministic (timing-sensitive, FatFs/LFS state ordering across tests), and PR #531's environment differed from this one.
2. A subtle interaction with another test that runs before \`test_blob_autoload\` is leaving state behind.

\`test_persistent_lfs_store\` tests are passing here, which rules out the persistent_lfs side of the original PR #486 fixes.

## Next steps

- Reproduce in isolation: run only \`test_blob_autoload\` (skip the others) and confirm same set of failures.
- Bisect against the test ordering — if a single earlier test causes pollution, that test's cleanup (\`tearDown\`) is incomplete.
- If non-deterministic, the per-test cleanup work in #512 is the long-term fix.

## Priority

P2 — \`make test\` is unusable as a quality gate while these fail. Can ride alongside #512.

[...]

Issue #532, opened 2026-04-28
Hailo backend: support model hot-swap via CCW re-upload


johnjezl:
## Context

Split out from #232 (model hot-swap). The base feature there swaps registry entries for models whose weights live in CPU-visible pools, with the runtime relying on \`ModelHandle\` refcounting to keep old weights alive while in-flight inferences finish.

That design doesn't work directly for Hailo-loaded models because the weights aren't held in a CPU pool — they're DMA'd to the Hailo-8 device at \`load\` time via the CCW (config-channel-words) upload path (\`hailo_control_upload_ccw\` → \`context_switch_load\` in \`kernel/inference/inference_device_hailo.c\`). The CPU-side pool only carries the HEF blob and metadata; what actually executes inference is on-device.

#232's first cut will reject \`model swap\` against a Hailo-backed slot with a clear error so the registry-level mechanism stays simple and safe.

## What this ticket covers

Implement the Hailo-specific swap path so \`model swap <hailo-idx> <new.hef>\` actually works on pi-5-1.

## Sketch

1. Add a backend-aware swap entry point on the inference device vtable, e.g. \`inference_device->swap(slot, new_data, new_len)\`. CPU backends use the trivial implementation (memcpy + atomic registry replace, handled by #232's runtime logic). Hailo provides a real one.
2. Hailo \`swap\`:
   - Allocate a *staging* model slot (separate from the live one) and run the full \`context_switch_load\` flow for the new HEF — CCW upload, boundary desc-list allocation, \`OPEN_BOUNDARY\` IOVA bind, \`SET_NETWORK_GROUP_HEADER\`, \`SET_CONTEXT_INFO\` chain, ACTIVATION.
   - Once staging is configured and idle, fence any in-flight inference on the old slot (the \`run_lock\` per-slot is already serialized by the VDMA ABI, but the swap operation needs to acquire it for the cross-over).
   - Send \`CHANGE_CONTEXT_SWITCH_STATUS\` to swing the active context to the new slot's network-group, then publish the new slot in the registry.
   - Tear down the old slot's CCW + boundary descriptor lists + free the on-device resources.
3. Failure handling: if the staging load fails partway, roll back cleanly — the live slot must remain usable (this is the main reason for the staging-then-commit shape rather than in-place re-upload).

## Acceptance

- \`hailo load\` model A; start an inference task in a loop; \`hailo swap\` to model B mid-loop; verify in-flight inference completes against A and subsequent inferences run against B.
[...]

Issue #237, opened 2026-04-17
x86-64 test suite: 10 pre-existing failures on make test PLATFORM=X86_64


johnjezl:
## Summary

`make test PLATFORM=X86_64` currently fails with **10 test failures**, all pre-existing on `main`. They are not regressions from any recent PR — they've been observed during build-verify on PRs #230, #236, and the rebase onto post-#235 main. Filing this as a single tracker so individual PRs stop needing to enumerate "failures unchanged, all baseline" in their test plans.

### Group 1 — PMM-exhaustion / model memory (3) — new on main after PR #235 (boot-preload)

All three fail with the same upstream cause: `[FAIL] Model memory init failed: PMM allocation failed` logged during test setup.

1. `test_shell_cmd_model_pin_lifecycle`
2. `test_shell_cmd_model_preload`
3. `test_shell_cmd_model_preload_wait`

**Hypothesis:** the boot-preload work allocates a larger model memory region earlier in init. On x86-64's smaller default RAM budget (`QEMU_TEST_MEMORY`), this pushes the allocator past a threshold by the time these tests run. ARM64 passes the same tests — suggests either a size difference or a platform-specific allocator ordering issue.

### Group 2 — Long-standing baseline (7)

4. `test_vmm_detected_ram`
5. `test_pic_timer_unmasked`
6. `test_platform_defines`
7. `test_gic_enable_disable_timer`
8. `test_timer_get_frequency`
9. `test_pci_shell_command_registered`
10. `test_pci_multifunction_device`

These have been failing for weeks. They cover legacy subsystems (VMM detection, PIC init ordering, x86 platform `#define` guards, PCI shell wiring, PCI multifunction enumeration) that no recent PR has touched.

## Why file this now

Every networking PR on the x86-64 path has to write "9 failures on main, same on my branch — unrelated" into its test plan. A single issue that's referenced instead keeps PR descriptions focused on the change and makes it obvious which failures are owned by whom.

## Reproduction

```bash
make kernel-test-clean
make test PLATFORM=X86_64
```

Last verified: 2026-04-17 on commit `18f42bc` (top of `net-irq-x86`, rebased on main past #235).

## Expected outcomes

- Group 1: Either the x86-64 test memory budget needs bumping, or the model-memory init needs to degrade gracefully when PMM is tight. Low risk — the boot-preload code path is well-tested on ARM64.
[...]

Issue #232, opened 2026-04-17
Model hot-swap: atomically replace weights while inference is in-flight


johnjezl:
## Context

Split out from #37 (Multi-Model Management). The registry, LRU eviction, pinning, and shell commands are all landed. The remaining scope item — replacing a model's weights atomically while inference tasks may be using them — is a distinct feature that requires reference-counting the weight handle so in-flight inferences complete on the old weights before they're freed.

## What needs doing

1. `rust_model_hot_swap(index, new_data, new_len)` — loads the new model into a staging slot, then atomically swaps the weight handle and graph under the registry lock.
2. In-flight inferences that already obtained the old weight pointer via `get_weights()` must complete safely. The old handle's refcount prevents premature free; the swap only replaces the registry entry.
3. Shell command: `model swap <index> <path>`.
4. Lua binding: `slm.model_swap(index, path)`.
5. Test: load model A, start a long-running inference task, swap to model B mid-flight, verify both the in-flight inference completes and subsequent inferences use model B.

## Acceptance

- Swap completes without crashing in-flight inference
- Old weights freed only after last reference drops
- Shell + Lua bindings work
- Test verifies concurrent safety

## References

- `runtime/src/loader/registry.rs` — registry with `ModelHandle` weight references
- `runtime/src/mm/model_mem.rs` — refcounted `share()` / `free()` on ModelHandle
- Component-level hot-swap already exists (`slm.component_hot_swap`) but operates at the task level, not the weight level

johnjezl:
Implementation plan written up: [docs/model-hot-swap-plan.md](https://github.com/SLM-OS/SLM-Operating-System/blob/main/docs/model-hot-swap-plan.md).

**Key points:**

- Pre-flight finding: the inference path (\`engine.rs\`, \`gpu.rs\`) currently calls \`get_weights\` (no refcount bump) — fixing it to use \`share_weights\` / \`unshare\` is a foundational prerequisite, not an add-on. There's a latent use-after-free against \`unload\` today even without swap.
- 7-step plan: refcount inference path → backend-aware dispatch → \`swap_model\` impl → FFI + shell command (\`model swap\`) → Lua binding (\`slm.model_swap\`) → tests (incl. concurrent-safety test that loops share/unshare while the main task swaps mid-flight) → docs.
[...]

PR #548, opened 2026-04-28
Implement model hot-swap (#232) and unblock x86-64 build

johnjezl:
## Summary

Implements the CPU-pool path of [#232](https://github.com/SLM-OS/SLM-Operating-System/issues/232) — atomic weight replacement for already-loaded models, safe under concurrent inference. Hailo CCW re-upload swap remains split out to [#532](https://github.com/SLM-OS/SLM-Operating-System/issues/532).

- **`registry::WeightLease`** (RAII, `runtime/src/loader/registry.rs:651`) — `share_weights` on construct / `unshare` on drop. Every `run_inference`, `gpu_map_weights`, and `gpu_unmap_weights` now holds a lease for its duration, so a swap mid-flight pins the OLD weight block until every reader drops; only then does the allocator deallocate.
- **`registry::swap_model`** (`runtime/src/loader/registry.rs:425`) — parse + alloc happen *outside* the registry lock (multi-MB work), atomic replace happens *inside*, OLD `mm::free` runs *after* the lock release. Refcount keeps the old block alive for any in-flight lease.
- **Backend dispatch** — `BackendKind` enum (`CpuOnnx | Hailo | Gpu`) on every slot. Non-`CpuOnnx` short-circuits to `LoadError::SwapNotSupported`; the C-side maps that to a friendly `model swap: ... (Hailo CCW re-upload tracked in #532)` message.
- **`prepare_onnx_payload`** helper extracted from `load_model` so both the load and swap paths share parser/allocator/populate logic.
- **Surface bindings:** `rust_model_swap` FFI, `model swap <name|idx> <path>` shell subcommand, `slm.model_swap(index, path[, name])` Lua binding.

### Drive-by fixes (unrelated pre-existing breakage)

- **x86-64 link break** — `kernel/mm/pmm.c` and `kernel/src/main.c` call `dtb_get_memreserves` / `dtb_get_chosen` unconditionally, but `kernel/src/dtb.c` is excluded on x86-64. Added `kernel/src/dtb_x86_stub.c` returning 'no entries'; platform-neutral callers handle that case cleanly. Detail in [#237](https://github.com/SLM-OS/SLM-Operating-System/issues/237).
- **Panic message truncation** — `panic_handler` file-path buffer bumped 64 → 256 bytes. The 64-byte buffer cut cargo registry paths exactly at the registry hash prefix (63 chars), hiding which crate panicked. Now the full crate-name + version + source file appears.

### QEMU
- [x] `make kernel PLATFORM=QEMU_VIRT` — clean
- [x] `make test` — `test_suite_model_swap` passes (Rust self-test + C-side FFI tests). Pre-existing flake on `test_blob_autoload` filed as [#533](https://github.com/SLM-OS/SLM-Operating-System/issues/533); model-swap suite passes on every run regardless.
- [x] `make kernel PLATFORM=X86_64` — was broken on `main` with undefined-reference errors; now clean.
- [x] `make test PLATFORM=X86_64` — now reaches a much later test (eviction demo) before hitting a pre-existing Rust panic in `linked_list_allocator-0.10.6/src/hole.rs:501` (`Freed node aliases existing hole! Bad free?`); details on [#237](https://github.com/SLM-OS/SLM-Operating-System/issues/237).

### Hardware
- [x] **jetson-nano-1** — kexec to SLM-OS, `model swap` shell subcommand wired (help text + all reject paths: empty slot → -2, missing file → -1, parse error → -4), 50-iter `model bench` clean (~4179 µs avg, σ negligible) — exercises `WeightLease` acquire/drop on every iteration.
- [x] **pi-5-1** — sdwire deploy, same battery, 100-iter bench clean (~3014 µs avg, σ ≈ 2 µs) — DSU coherency holds under repeated lease churn.
- [ ] x86-64 hardware — skipped per agreement; QEMU build now clean is the proxy.

### Coverage that QEMU owns

The successful end-to-end swap (load A → swap to B → verify pre-swap pointer still valid via the lease) runs entirely Rust-side via `rust_model_swap_test()` against the embedded MNIST. Couldn't be reproduced on real hardware because `/mnt/files/models/` is empty in the boot ramdisk and embedding a second ONNX would have been a separate plumbing change.

## Files

```
 CMakeLists.txt                  |   5 +
 docs/lua.md                     |   1 +
 docs/model-hot-swap-plan.md     |   2 +
 docs/specs/lua.md               |   2 +-
 kernel/include/slm_ffi.h        |  29 ++
 kernel/src/dtb_x86_stub.c       |  35 ++ (new)
 kernel/src/lua_slm.c            |  77 +++
 kernel/src/shell.c              |   2 +-
 kernel/src/shell_sys.c          | 112 +++++++-
 kernel/tests/test_harness.c     |   7 +
 kernel/tests/test_harness.h     |   3 +
 kernel/tests/test_model_swap.c  |  98 ++++++ (new)
 runtime/src/inference/engine.rs |  13 ++
 runtime/src/inference/gpu.rs    |  14 +-
 runtime/src/lib.rs              | 240 +++++++++++++-
 runtime/src/loader/registry.rs  | 333 +++++++++++++++++---
 runtime/src/mm/model_loader.rs  |   8 +
```


johnjezl:
Addressed review feedback in `ea9afcb`:

**Warnings (fixed):**
- W1 — Bounds-before-deref in `rust_model_swap` and (drive-by) `rust_model_load`. Conditions swapped so `name_len < 31` short-circuits the deref.
- W2 — `WeightLease` doc updated; the "Send but not Sync" claim was incorrect since both auto-impl. Sync is actually sound here.

**Suggestions (fixed):**
- S1 — Comment in `rust_model_swap_test` documenting the weights-only lease; explains why workspace doesn't need a parallel lease today.
- S2 — Comment in `swap_model` documenting that `backend_kind` is effectively immutable post-load, so the lock-drop between pre-flight and atomic swap is race-free.
- S5 — `model_swap` shell rc=-1 message reworded to flag it as a kernel bug (unreachable from a well-behaved caller).
- S6 — Comment block in `test_swap_empty_slot_after_init` warning about global registry mutation and noting which suite ordering it currently relies on.

**Deferred to follow-up #560:**
- S3 — Flatten `load_model` slot-find/install nesting (extract `find_or_evict_slot` + `install_payload`).
- S4 — Dedupe the file-read prologue across shell/Lua `model_load` + `model_swap`.

Both are pure cleanups with no behavioral change; the reviewer marked them as "don't block this PR" and they pair naturally as a single follow-up PR.

QEMU build clean; `make test` shows the `test_suite_model_swap` (14 Rust + 5 C-side) passing on every run, modulo the unrelated #533 `test_blob_autoload` flake.

Issue #537, opened 2026-04-28
shell-tcp leak: per-session heap delta still ~2920 B post-#442 fix (above 1024 B threshold)


johnjezl:
## Summary

Issue #442 closed 2026-04-27 with the claim that per-session lwIP heap leaks of ~1-2 KB had been addressed. Field observation on jetson-nano-2 (gpu-for-real branch, commit 30ebfaf) shows the leak warning still firing at session close:

```
[WARN] shell-tcp: session 2 closed with +2920 bytes still on lwIP heap
       (threshold=1024, measured post-tcp_close) — suspect leak
```

2920 B > the 1024 B threshold #442 considered acceptable, so either the fix missed paths or there's a new leak path in the upload (`xput`/`put`) flow.

## Why this matters

Sustained sessions exhaust the 128 KB lwIP heap. Once exhausted, `tcp_write` returns `ERR_MEM`, the shell-tcp drain stalls, and `tcp_write_buf` hangs forever (#536). User-visible symptom: telnet shell commands wedge the session.

PR #440 already raised `MEM_SIZE` 32 KB → 128 KB; another bump just delays the failure. Need to fix the actual leak.

## Repro

Same as #442:

```python
for _ in range(20):
    s = socket.create_connection(("192.168.4.5", 2323))
    time.sleep(0.4); s.close()
    time.sleep(0.2)
```

Then `netstat` to check `leaks: warnings=N total=M`. Field observation also shows the leak after a single session that did file uploads (`slm-put.py`, `slm-put-image.py`) with the put/xput path; need to isolate whether re-upload to existing path is a separate trigger from a clean close.

## Likely culprits #442 listed but didn't conclusively close out

1. `tcp_write(... TCP_WRITE_FLAG_COPY)` queued data not freed on abrupt RST or peer half-close mid-write.
2. `tcp_pcb->refused_data` after peer abandons before `tcp_recved` consumes everything.
3. Per-session `tcp_shell_ctx` allocation lifecycle.
4. RX-ring → pbuf hand-off ref-count off-by-one.

## Plan

1. Re-add per-allocation-type instrumentation in `kernel/src/shell_io_tcp.c` and `kernel/src/tcp_shell_server.c` so we can attribute the +2920 B to a specific allocation site.
2. Run the 20-cycle repro and the upload-then-close repro.
3. Fix the dominant leak; verify deltas drop ≤ 256 B (TIME_WAIT footprint expected).

## See also

[...]

PR #553, opened 2026-04-28
shell-tcp: bound write timeout, deferred leak measurement, leak attribution + gpu debug toggle

### LINKED ISSUE #536: shell-tcp: tcp_write_buf busy-loops forever when lwIP heap is exhausted
## Summary

`tcp_write_buf` in `kernel/src/shell_io_tcp.c:313` busy-loops with `sleep_ms` waiting for TX-ring space. When lwIP heap is exhausted (#442 leak path or a transient burst), the drain in `shell_io_tcp_poll` can't push pbufs out, the ring stays full, and the shell command hangs indefinitely.

User-visible symptom: any telnet shell command that produces multi-line output (`model stats`, `model gpu`, `tasks`, `mem`, etc.) freezes the session. Serial console is unaffected.

## Repro

On jetson-nano-2 over telnet (192.168.4.5:2323):

1. Open and close 10–20 telnet sessions to leak heap up against the watermark (or follow #442 repro).
2. Open a fresh session.
3. Run `model stats`, `model gpu`, or `tasks`.
4. Output stops after some number of lines; cursor never returns to `slmos>`.

A `[WARN] net: RX stalled (...) heap=131072/131072` warning typically fires alongside.

## Root cause

`kernel/src/shell_io_tcp.c:331-339`:

```c
irq_flags_t flags = spin_lock_irqsave(&ctx->tx_lock);
uint32_t avail = ring_free(ctx->tx_head, ctx->tx_tail);

if (avail == 0) {
    spin_unlock_irqrestore(&ctx->tx_lock, flags);
    /* Yield so net_pump can drain the buffer. */
    sleep_ms(TCP_SHELL_POLL_INTERVAL_MS);
    continue;
}
```

There is no upper bound on the retry count. If the drain loop in `shell_io_tcp_poll` (line 689-704) keeps getting `ERR_MEM` from `tcp_write` because lwIP heap is full, the ring never drains and the shell-task `while (src < len)` loop spins forever.

## Proposed fix

Bound the wait. Two options that compose cleanly:

1. **Total wall-clock cap** on the write call: e.g. ~2 s. After that, treat the write as failed, drop the rest of the buffer for this command, mark the session as degraded so future writes return immediately, and let the shell command return an error.
2. **Per-iteration cap** on `sleep_ms` retries: e.g. 200 retries × 10 ms = 2 s, then bail.

Either gets `model stats` printing `Command returned error: -1` instead of wedging the box. Pair with a one-shot `[WARN] shell-tcp: write timeout, dropping N bytes, marking session degraded` log so the operator can see what happened.

A degraded session would still let the user type `exit` or run commands that don't produce output, and the leak detector can still report on close. Or the bail can also tear the session down — both are reasonable; user input is uncommon under heap exhaustion anyway.

## Scope of fix

- `tcp_write_buf` in `kernel/src/shell_io_tcp.c` — main change.
- Add a `volatile bool degraded` (or repurpose `closed`) on `struct tcp_shell_ctx` so subsequent writes short-circuit.
- Regression test in `kernel/tests/test_shell_io_tcp.c` (or add a unit test if no fixture exists) that simulates a stuck drain and asserts `tcp_write_buf` returns within a bounded time.

## Not in scope

The actual heap-leak fix (per-session +2920 B above the 1024 B threshold) is a separate issue — see follow-up to #442. This issue is purely about the shell hang under heap pressure, not the leak itself.

johnjezl:
## Summary
Fixes #536, #537. Three commits on this branch:

- `4134e0a` — Bound `tcp_write_buf` retry at 2 s wall-clock, mark session degraded + closed instead of busy-looping forever when the lwIP heap is exhausted (#536). Defer the post-close heap-delta measurement by 1500 ms so unacked TCP_WRITE_FLAG_COPY pbufs have time to drain — eliminates the false-positive leaks #442 was chasing (#537).
- `c6622eb` — Snapshot `lwip_stats.memp[i]->used` at session open, diff at close. When a leak warning fires, the WARN line now appends `PBUF_POOL+2 TCP_SEG+1 …` so the operator can see which lwIP allocator the residual heap delta came from. Field repro on jetson-nano-2 mid-upload showed 2 of 13 closes still tripping the threshold post-fix; this attribution lands so the next investigation has the data it needs.
- `0bedd38` — Off-scope but small: gate the per-op `[GA10B-P8]` GPU dispatch trace behind a runtime flag (`gpu debug [on|off|status]`), default OFF. At 2 inf/s steady-state the trace was producing ~64 lines per inference and drowning the console. Errors and one-shot summaries bypass the gate. Happy to split this into its own PR if the reviewer prefers.

## Test plan
- [x] `make test` — `[PASS] test_shell_io_tcp_write_buf_timeout`, `[PASS] test_shell_io_tcp_close_settling`, `[PASS] test_tcp_shell_server_*`. The 12 pre-existing `blob_autoload` failures are present on main, unchanged.
- [x] Field repro on jetson-nano-2: `slm-put-image.py` upload — leak warnings now sporadic (2 of 13 sessions) instead of every-session; deltas attributable per attribution suffix.
- [x] Jetson hardware verification of the GPU debug toggle: `gpu debug status` → OFF default, `gpu debug on` → trace returns, `gpu debug off` → silent again, `model bench 1 1` runs cleanly with default OFF.


johnjezl:
Addressed actionable items from slmos-review. PR head is now `6df2d37`. Force-pushed to split the previous `fbb3bd3` commit into two cleanly-scoped commits + add a follow-up hygiene commit.

## Changes

1. **`fbb3bd3` split**:
   - `da40215` — gate pipeline-mode/complete summary lines (the change the original commit message described)
   - `90975b9` — reject GA10B handoffs/submits with >40-bit GPU VAs (the defensive validation work that was mistakenly bundled in)
   Future bisects across "GPU diagnostic output" vs "GPU VA validation" now resolve to distinct commits.

2. **`#include "uart.h"`** moved from line ~395 (mid-file) to the include block at the top of `shell_io_tcp.c`.

3. **Duplicated timeout block in `tcp_write_buf`** — extracted into `tcp_write_timeout_bail(ctx, dropped)`. Two ~17-line clones replaced with one-line calls.

4. **`shell_io_tcp_test_run_close_settling` cleanup symmetry** — restructured around a single `goto out` cleanup point with an idempotent `ctx_free` call. All three exit paths (success, scenario-1 failure, scenario-2 failure) now go through the same teardown.

5. **`memp_used_at_open[]` width bug** — the static_assert I added per the review caught a real (latent) truncation: `mem_size_t` is `u32_t` on our build because `MEM_SIZE = 131072 > 64000` (PR #440). The original `uint16_t` snapshot was silently truncating each `lwip_stats.memp[i]->used` reading on every session open. Pool used-counts are tens at most so no observed harm, but a real bug. Widened the snapshot to `uint32_t` (256 B more total) and pinned the assertion at `<= sizeof(uint32_t)` so a future widening of `mem_size_t` breaks the compile loudly.

## Test plan

`make test` — all 5 net suite tests still pass (including the two #536/#537 fixtures). 12 pre-existing `blob_autoload` failures unchanged from main. No new failures introduced.

Suggestions deferred:
- The `volatile bool` triumvirate (`closed`, `shell_done`, `write_degraded`) → enum state-machine refactor — too invasive for a follow-up to this PR. Can be addressed in a dedicated cleanup later.
- `kernel/src/shell_sys.c` `extern` decls inside `cmd_gpu` — also too invasive for this PR; would want a `shell_sys_jetson.c` split.

Issue #544, opened 2026-04-28
kernel: default_lconv.decimal_point overwritten at runtime — surface the writer


johnjezl:
## Summary

A page-fault report from a Lua `string.format("%.4f", ...)` call traced to `default_lconv.decimal_point` being overwritten at runtime with a bogus pointer (`0xffffffff8193c091`). The struct's ELF-initial value is correct (`0x801bf450`, pointing at `"."` in `.rodata`), so something writes to the static between boot and the first FP-format-via-Lua call.

`localeconv()->decimal_point[0]` then dereferences the bogus pointer → translation fault, level 1, kernel halt.

## Repro (ad-hoc)

Observed once on jetson-nano-2 running a longer Lua script (mnist_loop.lua w/ tostring + `%.4f`-driven workaround), kernel build off `gpu-for-real` / commit `30ebfaf`. Not yet a reliable repro; surfaces under sustained Lua scripting that exercises both `string.format` and other Lua paths.

## Mitigations already in place

The commit landing #536/#537 fixes (`fix/lua-printf-float`) makes the symptom go away even if the corruption persists:

- `default_lconv` moved to `.rodata` (now `const`). A repeat of the same write would fault on the write rather than silently corrupt — pointing the debugger directly at the culprit.
- `lua_getlocaledecpoint()` pinned to `'.'` (ASCII 46) via a CMake `-D` for the Lua target. Lua's float-format path no longer derefs `default_lconv` at all.

Either of those alone would mask the symptom; together they remove the user-visible page fault entirely.

## What's left

The underlying writer is unknown. Candidate questions to investigate:

1. With `default_lconv` now `const` in `.rodata`, can we reproduce the failure by exercising the same Lua workload? An access-violation fault during the write would identify the call-stack of the bad write directly.
2. Is anything memcpy/memset-ing into `.data` near `0x806a06c8`? Adjacent symbols are `udp_port` (16-bit), `tcp_port` (16-bit), `iss.0` (32-bit), `g_next_session_id` — small lwIP-ish counters. A buffer overrun starting in any of those would land here.
3. Is anything using a pointer past its declared size — e.g., the dual `struct lconv` definitions in `kernel/include/slm_lua_stubs.h` (2 fields) vs `kernel/src/lua_stubs.c` (18 fields)? The header form is currently unused in writes, but worth auditing for stale or out-of-sync declarations elsewhere.
[...]

Issue #538, opened 2026-04-28
SLM M5.3: real forward_step (registry tensor data + per-session tokenizer + weight pool)


johnjezl:
Umbrella: #475

The M5.2 `forward_step` (`runtime/src/slm/decoder.rs`) is a documented stub returning all-zero logits. Real text generation requires three plumbing pieces:

1. **Registry tensor-data ownership.** Today `runtime/src/slm/registry.rs` only stores `ArchInfo` + counts. Extend the entry to keep the parsed GGUF buffer alive (memory-mapped or copied into the weight pool) so the decoder can borrow tensor slices.
2. **Per-session tokenizer.** `Session::open` should construct + own the `Bbpe` tokenizer (currently the FFI uses a single-token placeholder). The non-empty-prompt path of `rust_slm_prompt` returns -1 today; M5.3 makes it work.
3. **forward_step over the M4 op chain.** Walk: embedding → for each layer (rmsnorm → gqa_decode_step → rmsnorm → swiglu_mlp → residual) → final rmsnorm → lm_head → logits. Use the M3 vec_dot helpers via M4's `matmul_q4k_*`. Append KV at every layer; commit position once per token.

Acceptance: `slm prompt 0 "What is virtual memory?"` on Jetson produces coherent text.

References:
- M4 ops: `runtime/src/inference/ops_transformer.rs`
- M5.2 stub: `runtime/src/slm/decoder.rs::forward_step`
- Registry: `runtime/src/slm/registry.rs`

johnjezl:
M5.3.1 (#545) + M5.3.2 (#547) merged via #549. 

The real forward pass is now wired end-to-end:
- Registry owns the GGUF byte buffer in the weight pool, builds a per-model Bbpe tokenizer, snapshots all tensor descriptors.
- runtime/src/slm/forward.rs::forward_one walks the full Qwen2 pipeline: Q4_K embedding → per-layer (RMSNorm + QKV matmul w/ Qwen biases + RoPE + KV-append + GQA + output proj + residual + RMSNorm + SwiGLU + residual) → final RMSNorm → LM head.
- ForwardScratch owns the per-token buffers; decode hot loop is allocation-free.
- The KV-view ordering bug (gqa was called with pre-commit view, missing the just-appended slot) caught + fixed during review.

152/152 runtime tests passing.

[...]

PR #554, opened 2026-04-28
kernel,lua: freestanding %f/%g/%e + locale-deref bypass + strpbrk prototype

johnjezl:
## Summary

Three diagnostic-output fixes traced to a single Lua page-fault report (`#544`, filed alongside).

- `5a90442` — Pin `lua_getlocaledecpoint()` to literal `'.'` (ASCII 46) via a CMake compile option for the Lua target. Bypasses the deref of `localeconv()->decimal_point[0]` that was page-faulting after runtime corruption of `default_lconv` (separately tracked in #544). Also makes `default_lconv` `const` so it lands in `.rodata` — any future write attempt now faults at the writer rather than silently corrupting `.data`. Adds full `%f` / `%F` / `%e` / `%E` / `%g` / `%G` support with `.N` precision parsing — Lua's `string.format("%.4f", x)` previously emitted the format string verbatim because our printf stub lacked float specifiers. Float-formatting code lives in a new `kernel/src/kprintf_float.c` static library compiled WITHOUT `-mgeneral-regs-only` so it can do real `double` arithmetic; the rest of the kernel stays under the FP-register ban. `kprintf.c` passes a `va_list*` across the boundary so it never references `double` itself.
- `b30a4a5` — Trivial: correct the issue ref in CMakeLists.txt (#538 → #544 once the actual issue number was assigned).
- `d3f2bc6` — Add `strpbrk` / `strrchr` / `strstr` prototypes to `kernel/include/string.h`. Lua's `lobject.c:luaO_str2num` calls `strpbrk` without an in-scope prototype; the compiler defaulted to implicit-int return and emitted `sxtw x21, w0` to sign-extend the truncated 32-bit return. The result was a `0xffffffff8XXXXXXX`-shaped pointer that translation-faulted on the next deref. Surfaced after the locale bypass landed (the previous fault site was masked). The implementations already existed in `kernel/src/lua_stubs.c` and `kernel/src/string.c`; only the declaration was missing.

## Test plan
- [x] `make test` — 6 new float-format tests pass (`test_kprintf_float_*`, `test_kprintf_scientific`, `test_kprintf_g_format`). Bench tests use a uint64-bits helper so test code stays under `-mgeneral-regs-only`. Pre-existing `blob_autoload` failures unchanged from main.
- [x] Jetson hardware verification: `lua /mnt/files/f.lua` running `print(string.format('%.4f', 3.14159))` returns `3.1416` (was page-faulting). Broader smoke test exercises `%f`/`%e`/`%g` across Inf, NaN, and normal values.


johnjezl:
Addressed all actionable items from the slmos-review pass. PR head is now `e1a9bbe`. Summary follows; full details in the commit message.

## Warnings — all 3 addressed

**W1 (x86-64 build)**: `make kernel PLATFORM=X86_64` now passes. Surfaced a real GCC `-Werror=incompatible-pointer-types` issue along the way: x86-64 SysV `va_list` is `__va_list_tag[1]` (array), so `&args` is `T(*)[1]` not `T**`. Added `(va_list *)&args` cast in `kprintf.c` with a comment explaining the array-vs-pointer ABI quirk. `make test PLATFORM=X86_64` builds; the runtime crash is a pre-existing Rust panic on main (verified by stash + retest of plain main).

**W2 (end-to-end va_list test)**: new `test_kprintf_float_e2e_va_list_crossing` routes through the production path (`snprintf` wrapper in `lua_stubs.c` — the same chain Lua's `string.format("%.4f", x)` uses). Three assertions covering `%.4f`, `%g`, and `%e` of literal doubles, all passing.

While writing this test we surfaced (and verified) that `uart_snprintf` itself can't reliably handle FP args because its `va_start` runs under `-mgeneral-regs-only`. Production isn't affected because no caller passes FP args directly to `uart_snprintf`; Lua's path goes through the FP-enabled `lua_stubs.c::snprintf` wrapper. Filed separately as **#562**.

**W3 (padding-branch coverage)**: split out `kprintf_float_emit_value` so the test seam `kprintf_float_test_format_bits_padded` can drive the padding logic without faking a va_list. Five new tests covering: sign-then-zeros for negative, trailing spaces for left-justify, and the inf/-inf/nan space-pad-not-zero behaviour. The 0-flag-with-non-finite suggestion is now also enforced (we follow glibc/musl).

## Suggestions — 6 of 7 addressed

- **S1**: `-Dlua_getlocaledecpoint()=46` → `'.'` (character literal, equivalent under C 6.4.4.4 but readable).
- **S2**: comment on `(void)arg_is_64bit` noting `L` (long double) is intentionally unsupported.
- **S3**: `tmp[24]` rationale hoisted to a single `UNSIGNED_TO_DECIMAL_BUF_SIZE` named constant; reworded to drop the platform-specific x30 detail.
- **S4**: `1e18` / `1.8e19` promoted to `FIXED_PATH_MAX_VALUE` / `SCALED_CAST_LIMIT` named constants.
- **S5**: recursion-bound comment added — `fmt_fixed` ↔ `fmt_sci` is depth-1 because `sci` normalises to `[1, 10)` before re-entering `fixed`.
- **S6**: `DBL_BITS_*` table now carries a Python derivation comment (`struct.pack('>d', x)`) plus a note about why we don't trust printf round-tripping for these.
- **S7** (const-then-cast in `lua_stubs.c::localeconv`): deferred per the review's own "current shape is fine" guidance.

## Test plan

- `make test` (QEMU ARM64): all 12 float-format tests pass (existing 6 + 5 padded + 1 e2e).
- `make kernel PLATFORM=X86_64`: clean.
- `make test PLATFORM=X86_64`: kernel builds; runtime crash pre-existing on main.

Ready to merge.

Issue #535, opened 2026-04-28
Jetson: ai_mlp+gpu-use-sched dispatch wedges box under task burst (sync IRQ-off cliff)


johnjezl:
## Summary

With `sched policy ai_mlp` + `gpu use sched on`, every `ai_mlp_assign_cpu` call routes through `slm_gpu_run_sched_inference` synchronously while holding `g_gpu_dispatch_lock` (`spin_lock_irqsave`). Per-call cost is ≈ 5 ms IRQ-off vs ≈ 50 µs on the CPU NEON path. Even moderate task-creation cadence (the industrial IoT demo, a 2-iteration Lua loop) pins the kernel hard enough that telnet, serial console, and networking all stop responding — recovery requires a power-cycle.

A single `slm.task_create()` works cleanly: scanner finds the kind=1 handoff, all 8 ops dispatch through GPFIFO, control returns to the shell. The cliff appears only with sustained task creation.

PR-3 of `docs/specs/gpu-policy-models.md` introduced the dispatch path and `gpu_consumer.c` does emit a perf-note on toggle:

> "perf note — ai_mlp now dispatches each assign_cpu via GPU (~5 ms vs ~50 us CPU NEON); fallback to CPU on dispatch error"

But the warning describes the per-call latency only; there's no actual cliff handler — there's no rate limit, no fallback when dispatches queue up, and no async path. With six CPUs all potentially entering `assign_cpu` from concurrent task-create / work-stealing paths and serializing on `g_gpu_dispatch_lock`, the IRQ-off accumulation starves the timer tick, the lwIP RX-stall watchdog, and shell scheduling.

## Repro (jetson-nano-2, fresh kexec from `gpu-for-real` branch / commit `30ebfaf`)

1. `slmos-kexec --no-gpu-suspend /root/slmos.elf` (with both `gpu-kernel-mnist` and `gpu-kernel-sched-mlp` helpers preserved).
2. `sched policy ai_mlp`
3. `gpu use sched on` (perf note prints; toggle accepts).
4. Either:
   - `lua-admin /mnt/files/demo.lua`, or
   - a 2-iteration Lua loop calling `slm.task_create('t', function() end)`.

Expected: dispatch fires once per `assign_cpu`, system continues responding.
Actual: shell + telnet + ICMP all stop responding mid-run; ~`gpu use sched off` would have prevented it; only power-cycle recovers.

[...]

PR #555, opened 2026-04-28
gpu: consecutive-failure circuit breaker (#552 mitigation)

### LINKED ISSUE #552: Inference dispatch wedges shell after sustained run (~120 iterations)
## Summary

Running `mnist_loop.lua` (a tight loop of `slm.model_infer_file`) over telnet on jetson-nano-2 wedges the entire shell — both the telnet session AND the serial console — after ~120 successful GPU inferences. Power-cycle is the only recovery.

This is a separate failure mode from #535 (sched-MLP dispatch wedges box under task burst). The trigger here is sustained inference dispatches from a single shell task, not concurrent task creation across CPUs. But the underlying mechanism is suspected to be the same architectural class: synchronous IRQ-off GPU dispatch starving the system.

## Reproducer

1. Boot SLM-OS on jetson-nano-2 with the channel-inherit helper staged (any current `gpu-for-real` tip works).
2. `model load mnist; gpu use inference on; model use-gpu 1 on`.
3. Run a Lua script that loops `slm.model_infer_file(idx, path)` indefinitely.
4. After ~120 successful iterations, the shell stops responding.

User-visible signature on the wedge:

```
[GA10B-P8] pipeline mode — 8 ops, ops_phys=0x13e5f5000
[GA10B-P8] pipeline complete — 8 ops fired
[                                          <-- next dispatch starts, never finishes
```

(The `[GA10B-P8]` lines are gated behind `gpu debug on|off` in the latest tip; reproduction with debug off would just stop printing inference results without the explanatory trace.)

## Suspected root cause

`kernel/gpu/nvidia/ga10b_bringup.c:ga10b_submit_and_poll` busy-waits on the semaphore for up to **2 seconds** with the GPU dispatch lock held IRQ-off (acquired at the FFI boundary in `kernel/src/slm_ffi.c:slm_gpu_run_mnist`). Per inference: 8 ops × successful poll ≈ a few ms each, total ~10 ms IRQ-off — fine.

If at iteration N the GPU stops accepting dispatches (GPFIFO ring near-wrap, FECS state drift, anything), each op polls the full 2 s before failing. 8 ops × 2 s = **16 s of IRQ-off polling** per failed inference. Timer ticks, scheduler, network pump, watchdog all starve during that window. By the second or third such failure, the system is hung even though the GPU dispatch itself eventually returns -1 to the caller.

The 120-iteration mark is suspicious because the GPFIFO ring has 1024 entries and 120 × 8 = 960 — close enough to the wrap point that ring-wrap handling deserves a careful look. (`g_handoff.initial_gp_put` is monotonic in our code, indexed via `& (gpfifo_entries - 1)` for the slot — that part looks correct, but the channel state on the GPU side might not handle the wrap as we expect.)

## Why #535's perf-note doesn't cover this

#535 is about `ai_mlp_assign_cpu` GPU dispatch on every task creation. This issue is about `model.infer` GPU dispatch on every Lua call. The shared underlying problem is the **synchronous-IRQ-off-during-poll** pattern; both consumers exhibit it but from different call paths.

## Proposed fix (per #535's option A, applied to inference too)

Move the per-op poll out of the IRQ-off critical section:
- The dispatch lock is needed to serialize updates to `g_handoff.initial_gp_put / initial_gp_get` and the GPFIFO ring head — pushbuf and doorbell write.
- The poll itself only reads the semaphore and the GP_GET counter; it does NOT need exclusive ownership of the channel state.
- Restructure `ga10b_submit_and_poll`: hold the lock for push + doorbell, drop it, then poll without IRQs disabled, then reacquire briefly to update the bookkeeping if successful.

Net effect: a wedged 2-s poll no longer starves the rest of the system. The polling task can be preempted by timer ticks during the wait, and the rest of the kernel keeps running.

## Acceptance

- 1000-iteration mnist_loop run completes without wedging the shell.
- A deliberately-wedged GPU (e.g., test path that makes poll always fail) returns -1 to the caller within ~2 s without making the system unresponsive.
- Existing single-shot dispatches still work end-to-end.

## Workaround until then

Limit sustained inference to <100 iterations per shell session, or drive inference from the Lua REPL with manual pacing.

## See also

- #535 — same failure pattern from the sched-MLP consumer
- `kernel/gpu/nvidia/ga10b_bringup.c:ga10b_submit_and_poll`
- `kernel/src/slm_ffi.c:slm_gpu_run_mnist` — the IRQ-off lock holder

johnjezl:
## Summary

Mitigation for #552 — GPU dispatch wedges shell after sustained inference. **Not** the architectural fix (that needs the busy-poll moved out of the IRQ-off section); this is the small, standalone protective patch that gets the system out of the 2-s-per-iteration tax once the GPU enters a degraded state.

### What's the failure mode

After ~120 successful dispatches, the GA10B GR engine on jetson-nano-2 stops producing the post-launch semaphore release (likely GPFIFO ring-wrap aftermath / channel state drift — see #552 for diagnostics). PBDMA still accepts the submit (GP_GET advances), but `ga10b_submit_and_poll` busy-waits the full 2 s timeout for a payload that never arrives, returns -1, engine.rs falls back to CPU. Every subsequent dispatch hits the same path.

Each failed dispatch holds `g_gpu_dispatch_lock` IRQ-off on CPU 0 for those 2 s. With 10 inferences/pass × 2 s = 20 s of IRQ-off per script iteration on CPU 0, `net_pump` can't service telnet, `Ctrl+C` IAC processing on the same CPU is starved, and the shell appears wedged from the operator's POV.

### What this PR does

Tracks consecutive failures of `slm_gpu_run_mnist` / `slm_gpu_run_sched_inference` in a shared atomic counter. After 3 back-to-back failures both entry points short-circuit to `-1` **before taking the IRQ-off lock** — `engine.rs` already falls back to CPU NEON on -1, so the user-visible result is unchanged but the polling cliff is eliminated.

- One-shot WARN exactly when the threshold crosses; subsequent failures don't flood the log.
- Counter resets on any successful dispatch (forward-looking — useful once #552's recovery path lands).
- `gpu use inference on` clears the breaker so the operator can manually retry after a power-cycle / kexec.
- Stubbed cleanly on non-Jetson.
- Threshold of **3** is high enough that a single transient timeout doesn't disable the fastpath, low enough that a real wedge is contained within ~6 s of total IRQ-off.

### Why not just fix #552 directly

The architectural fix is to drop the dispatch lock during the poll phase — only re-acquire briefly for bookkeeping. That's substantially more invasive (touches the channel-state ownership model) and worth doing carefully. The breaker is small enough to land standalone and gives operators their shell back **today** while the deeper work is in flight.

## Test plan
- [x] `make test` (QEMU): kprintf-float + shell-tcp suites still green; pre-existing `blob_autoload` failures unchanged from main.
- [ ] Jetson hardware: deploy, re-run `mnist_loop.lua`, confirm the breaker trips after the GPU degrades and the system stays responsive (no per-iteration 2 s pause once tripped). I'll add results in a comment after deploy.

## See also

- #552 — root cause + architectural fix (separate work)
- #535 — same IRQ-off-cliff pattern from the sched-MLP consumer


johnjezl:
Round-1 /slmos-review response (commit 2535d55).

**Warnings**
- **W1 (test coverage):** added `kernel/tests/test_gpu_dispatch_breaker.c` — 8 Jetson-only tests + 1 non-Jetson stub test. Wired into the harness via `CMakeLists.txt` + `test_harness.{c,h}`. Uses two new test seams (`slm_gpu_dispatch_breaker_test_record` / `_test_count`) declared in `slm_ffi.h` so the suite can drive the state machine without a real GA10B. Coverage: default state, single/two failures don't trip, three failures trip, success-resets-counter, manual-reset-clears-tripped, post-trip continues counting, intermittent-success-resets-then-restarts.
- **W2 (hardware verification):** in flight. Will deploy to nano-2 and report back with serial transcript showing WARN fires once at threshold + `gpu use inference on` reset path.

**Suggestions**
- **S1 named threshold:** `GPU_DISPATCH_BREAKER_THRESHOLD` is now `#ifndef`-overridable from CMake/Make.
- **S2 internal predicate rename:** `gpu_dispatch_breaker_tripped` → `gpu_dispatch_breaker_is_tripped_inner` (matches the public wrapper).
- **S3 memory_order_relaxed rationale:** added a paragraph above the atomic explaining why relaxed is correct here.
- **S4 issue refs:** dropped `(see #552)` from the WARN string; it'll bit-rot when the issue closes.
- **S5 shared-counter assumption:** documented why MNIST + sched-MLP share one counter (same channel, degrade together); future channel-split work needs to split the counter too.
- **S6 lock-order doc:** comment on `gpu_dispatch_record_result` explaining it must run *after* `spin_unlock_irqrestore` (uart_printf takes its own UART lock).

**Build status**
- QEMU `make test`: PASSED (new suite green via stub path).
- Jetson `make kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON`: PASSED.
- x86-64 `make kernel PLATFORM=X86_64`: pre-existing failure (`dtb_get_memreserves` / `dtb_get_chosen` multiple-definition on plain `origin/main`, unrelated to this PR).

johnjezl:
**W2 (hardware verification) — partial / deferred:**

I claimed jetson-nano-2 to deploy the breaker build and observe the WARN/INFO lines on real UART. Got as far as:

- Built `kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON` cleanly with the new code (md5 `717422747e18864adb098fb93198997e`).
- Copied the ELF to nano-2 over SSH and verified the md5 matches.
- Power-cycled, confirmed Linux boot, kicked off `slmos-kexec /root/slmos.elf` over SSH (default suspend mode).

**Blocker:** The labctl serial console (`/dev/lab/port-2-1`, alias `jetson-console`) returned `Port already in use` for every `serial_capture` and `serial_send` attempt before, during, and after the kexec. That's a labctl/ser2net infrastructure issue (probably a stale ser2net session holding the port), not anything to do with this PR. Without UART access I can't directly observe the breaker's `[WARN] gpu-dispatch:` line on real hardware.

**What's verified:**
- Breaker state machine: 8 unit tests pass via the test seams in `kernel/tests/test_gpu_dispatch_breaker.c` (default state, single/two/three failures, success-resets, manual-reset, post-trip continues counting, intermittent-success-resets-then-restarts).
- Production wiring: `slm_ffi.c:586` (run_mnist) and `slm_ffi.c:643` (run_sched_inference) both check `gpu_dispatch_breaker_is_tripped_inner()` before taking the dispatch lock; `gpu_dispatch_record_result(rc)` is called after every dispatch attempt with the tail-call comment about UART lock-order. The diff is identical to what unit tests exercise.
- Compilation: Jetson build is clean with all the new code path active.

**What's still pending:**
- Live observation of the `[WARN] gpu-dispatch: 3 consecutive failures — circuit breaker tripped` line firing once, exactly when the threshold is crossed, on real hardware.
- Live observation of `[INFO] gpu-dispatch: circuit breaker reset (was tripped at N consecutive failures)` after `gpu use inference on`.

I'll fold both into the next live serial session; not blocking merge from my side, but flagging so the W2 box doesn't get checked off without an actual UART transcript.

johnjezl:
**W2 (hardware verification) — COMPLETE on jetson-nano-2**

Live verification on real GA10B caught a real defect that round-1 review didn't surface: the breaker was wired into `slm_gpu_run_mnist` + `slm_gpu_run_sched_inference`, but engine.rs's MNIST fastpath calls `slm_gpu_set_mnist_input` *first*. When `ensure_mnist_bringup()` failed inside `set_mnist_input` (which is exactly what happens with a priv-locked channel — `HWCFG2=0x0001a733 bit13=1`), `run_mnist_gpu_fastpath` returned `Err(DispatchFailed)` immediately and `slm_gpu_run_mnist` was never reached. Counter never moved → breaker never tripped.

Fixed in commit d7e2c41 by mirroring the same breaker discipline in `slm_gpu_set_mnist_input` (and `_fill` for symmetry).

**Live transcript (jetson-nano-2, GPU forced into degraded state via default `slmos-kexec` GPU-suspend mode):**

```
[3179.73] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[GA10B-INHERIT] HWCFG2=0x0001a733 bit13=1
[3179.80] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 24333 us
[3239.65] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[GA10B-INHERIT] HWCFG2=0x0001a733 bit13=1
[3239.70] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 24323 us
[3299.63] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[GA10B-INHERIT] HWCFG2=0x0001a733 bit13=1
[WARN] gpu-dispatch: 3 consecutive failures — circuit breaker tripped, fastpath disabled until next successful dispatch or `gpu use inference on`
[3299.74] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 37355 us
[3360.61] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[3360.65] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 13552 us
[3419.70] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[3419.74] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 13546 us
[INFO] gpu-dispatch: circuit breaker reset (was tripped at 3 consecutive failures)
[3539.57] [INFO] [engine] mnist GPU fastpath eligible -- dispatching[GA10B-INHERIT] HWCFG2=0x0001a733 bit13=1
[3539.64] [INFO] [engine] mnist GPU fastpath failed -- falling back to CPUInference on 'mnist' completed in 24322 us
```

**Observed behaviors (every one matches the unit-test expectations):**

1. **Threshold trip:** `[WARN] gpu-dispatch: 3 consecutive failures — circuit breaker tripped, …` fires *exactly* at attempt 3, exactly once.
2. **IRQ-off short-circuit confirmed:** attempts 4 and 5 still run the engine eligibility check, but the `[GA10B-INHERIT]` probe line is **missing** — the breaker short-circuits the call before the IRQ-off lock + `ensure_mnist_bringup` probe. Visible in the timing: 24333 µs (with probe) → 13552 µs (CPU-only fallback). That's the entire point of the breaker.
3. **Manual reset:** `gpu use inference on` prints `[INFO] gpu-dispatch: circuit breaker reset (was tripped at 3 consecutive failures)` once, with the live count.
4. **Post-reset full re-probe:** attempt 6 shows the `[GA10B-INHERIT]` line back at 24322 µs — counter is at 0, breaker is open, full ensure_mnist_bringup probe re-engages.

**Healthy-GPU no-regression check** (separate run, `slmos-kexec --no-gpu-suspend`, channel-inherit succeeded `HWCFG2=0x00018733 bit13=0`):
- 6 successful dispatches in a row, counter stayed at 0, no spurious WARN, no spurious INFO from `gpu use inference on`. ✓

W2 done. Standing rule says no merge without explicit go-ahead — flagging for your call.

johnjezl:
Round-3 review response (commit 06ceeb3):

- **Filter caller-side errors** — added `gpu_dispatch_record_set_input_result` that only forwards -1 / 0 to the breaker counter. -2 (cap too large) and -3 (bad arg) are dropped entirely (neither increment nor reset). Applied at both `slm_gpu_set_mnist_input` and `_fill` call sites. Comments at both sites point to the helper for the rationale.
- **Test seams off the public header** — moved `_test_record` / `_test_count` declarations into a new `kernel/include/slm_ffi_test.h`. Production TUs that include `slm_ffi.h` can no longer reach them. Unit test file picks up both headers (added an inline note explaining the split).

The third round-3 suggestion (test output emits one-shot WARN/INFO lines during make test) was explicitly deferred as cosmetic — not addressed here.

QEMU `make test`: PASSED. Jetson `make kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON`: PASSED.

PR #556, opened 2026-04-28
slmos-kexec: auto-stage GPU helpers when --no-gpu-suspend

johnjezl:
## Summary

Default the GPU channel-inherit helper-staging step into `slmos-kexec` so deploys "just work" without the operator having to remember to launch `gpu-kernel-mnist --preserve-for-kexec` separately. Without an active channel, Linux's nvgpu leaves priv-lock asserted (HWCFG2 bit 13), SLM-OS's channel-inherit fails, and inference falls back to CPU.

## What changes

Two behaviours, both gated on `--no-gpu-suspend` (the GPU-suspend path power-cycles the engine and has no Linux state to inherit, so it doesn't need a helper):

- **Detect** whether `gpu-kernel-mnist` or `gpu-kernel-sched-mlp` is already running with `--preserve-for-kexec`; if so, leave it alone (the operator may have staged their own with custom weights / shaders).
- **Otherwise launch** the binaries from `$SLMOS_HELPER_DIR` (default `/root/gpu-mnist`) via `setsid + nohup`, wait up to 60 s for each to print "kexec now" (the helper's last log line before it sleeps), then continue. Sched-mlp is best-effort — if the binary or weights aren't on disk, we skip it without failing.

New `--no-helper` opt-out for CPU-only inference or out-of-band-staged helpers. `SLMOS_HELPER_DIR` env var documented for non-default layouts.

### Step-numbering change (consumers please re-check)

Step prefix bumped 1..7 → 1..9 to make room for the new `[4/9]` GPU helper-stage step and the `[6/9]` runtime-PM pin step that was previously folded into [5/7]. Anything that greps `slmos-kexec` output for a literal `[N/7]` string will need to update. Searched the repo + lab tooling — found no consumers; calling it out here in case private runbooks or operator dashboards do.

## Test plan
- [x] Verified live on jetson-nano-2: `slmos-kexec --no-gpu-suspend` auto-stages `gpu-kernel-mnist` + `gpu-kernel-sched-mlp` from `/root/gpu-mnist`, both reach "kexec now" within ~10 s, kexec proceeds. SLM-OS post-kexec: `HWCFG2=0x00018733 bit13=0` (priv-lock cleared), all 8 MNIST GPU ops fire with `payload observed`. Without this change, the same deploy left priv-lock asserted and every dispatch fell back to CPU.

### DIFF
--- a/scripts/jetson-kexec-slmos.sh
+++ b/scripts/jetson-kexec-slmos.sh
@@ -37,6 +37,7 @@
 #   sudo ./jetson-kexec-slmos.sh --no-usb-hold   /path/to/slmos.elf
 #   sudo ./jetson-kexec-slmos.sh --no-smmu-fix   /path/to/slmos.elf
 #   sudo ./jetson-kexec-slmos.sh --no-usb-root-cleanup /path/to/slmos.elf
+#   sudo ./jetson-kexec-slmos.sh --no-helper     /path/to/slmos.elf
 #
 # Or install to Jetson and run:
 #   sudo slmos-kexec /root/slmos.elf
@@ -71,12 +72,31 @@
 #                       cleaner addressed state and skips only the
 #                       downstream slot-3 handoff for that run.
 #
+#   --no-helper         Skip the auto-start of the GPU channel-inherit
+#                       helper. By default, when --no-gpu-suspend is set,
+#                       this script starts gpu-kernel-mnist (and, if
+#                       present, gpu-kernel-sched-mlp) with
+#                       --preserve-for-kexec from $SLMOS_HELPER_DIR
+#                       (default /root/gpu-mnist) before loading the new
+#                       kernel. Without an active GPU channel, Linux
+#                       boots leave priv-lock asserted (HWCFG2 bit 13),
+#                       SLM-OS's GPU fastpath fails the channel-inherit
+#                       check, and inference falls back to CPU. Use
+#                       --no-helper if you've staged your own helper or
+#                       deliberately want CPU-only inference.
+#
+# Environment:
+#   SLMOS_HELPER_DIR    Directory containing gpu-kernel-mnist /
+#                       gpu-kernel-sched-mlp + their weights/shaders.
+#                       Defaults to /root/gpu-mnist.
+#
 set -euo pipefail
 
 NO_GPU_SUSPEND=0
 NO_USB_HOLD=0
 NO_SMMU_FIX=0
 NO_USB_ROOT_CLEANUP=0
+NO_HELPER=0
 KERNEL=""
 SKIP_XHCI_SLOT3_HANDOFF=0
 XHCI_SLOT1_HANDOFF_PAYLOAD=""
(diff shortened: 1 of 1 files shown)


johnjezl:
Round-1 /slmos-review response (commit 80dabf3).

**Warning**
- **W1 (handoff contract stringly-typed):** added a comment near the `grep -q "kexec now"` site explaining that the string is a contract with `host-tools/gpu-kernel-{mnist,sched-mlp}/main.cpp` and must be updated in lockstep if the helper output changes.

**Suggestions**
- **S1 (silent cd failure):** added `[[ -d "$helper_dir" ]]` check before the `(cd … && nohup …)` subshell. `set -e` does not propagate into a backgrounded subshell, so silent cd failures previously surfaced as "helper started successfully but never logged anything" 60 seconds later.
- **S2 (`pgrep -f` regex):** switched to `pgrep -fx` for exact-match. Plain `-f` does a regex substring match against the joined argv. The `-fx` form anchors against the full command line. Two patterns are tried (absolute + `./`) because helpers are launched relative from inside `helper_dir`, but a previously-staged process may have used the absolute path.
- **S3 (`tail -10` → `tail -50`):** bumped log tail size in the timeout warning so operators have more context when a wedged helper fails to reach the handoff string.
- **S4 (/tmp logs accumulate):** **deferred via #563.** Logs are at fixed paths in /tmp (tmpfs) and truncated each run via `: > "$log"`, so accumulation is bounded. Issue #563 captures the case where `slmos-kexec` migrates to a long-lived host-side context that needs per-run rotation.
- **S5 (PR body note about renumbering):** PR description now has a "Step-numbering change (consumers please re-check)" section calling out the 1..7 → 1..9 bump explicitly. No consumers found in the repo or lab tooling.
- **S6 (distinct exit codes):** documented the three exit codes explicitly in `start_one_helper`'s header comment block: `0` newly started or already running, `1` skipped (binary/weights missing), `2` timeout. Both call sites still use `|| true` because every non-zero outcome is non-fatal today, but future callers can branch on the specific failure mode.
- **S7 (`--help` flag):** **deferred via #564.** Header comment block is sufficient documentation; adding `--help` would push the option-parsing block into a real getopt parser. Tracking issue captures the work.

**Test status**
- `bash -n scripts/jetson-kexec-slmos.sh`: PASSED
- `shellcheck -S warning`: clean (no warnings)
- Hardware-side: no functional change to the staging flow, only error reporting + matching hygiene; will exercise on next `slmos-kexec` deploy.

johnjezl:
Round-2 review response (commit ccc731d):

- **`pgrep -fx` regex escape:** added a sed-based escape pass for ERE specials (`. * + ? ( ) [ ] { } | \\ ^ $`) before substituting into the `pgrep` pattern. Smoke-tested against synthetic argv strings — a `SLMOS_HELPER_DIR` containing a literal dot now correctly fails to match a sibling path with that position substituted.
- **Failure-mode ordering:** moved the `[[ -d "$helper_dir" ]]` check to the top of `start_one_helper`, removed the redundant post-launch check. Order is now monotonic: dir → binary → weights → launch.

`bash -n` and `shellcheck -S warning` clean. No remaining suggestions; deferred items #563 and #564 stand from round-1.

PR #557, opened 2026-04-28
runtime: drop fastpath log_info chatter + route Rust prints via shell_printf

johnjezl:
## Summary

Two related runtime cleanups so steady-state inference doesn't spam the console and shell-driven output reaches the bound TCP session (telnet, not just UART).

### (1) `engine.rs`: drop per-iteration fastpath chatter

Remove the `[engine] mnist GPU fastpath eligible -- dispatching` and `[engine] mnist GPU fastpath success` `log_info` calls. They were debugging tracers from when the fastpath was being brought up; once it's working, every inference firing them is just noise. The `failed -- falling back to CPU` line **stays** because that's the operator-actionable case.

### (2) `lib.rs`: `uart_printf` → `shell_printf` (bulk rename, 64 sites)

User-facing Rust FFIs that print through the shell now actually reach the bound shell session:

- `rust_gpu_print_status` (`model gpu`)
- `rust_infer_bench` (`model bench`, latency block)
- `rust_infer_and_print` (`model infer`)
- `rust_infer_buf_and_print` (`model infer-file` logits dump)
- `print_logits_and_argmax` (helper)

Plus the boot/test prints that share the same extern. `shell_printf` already routes to the bound `shell_io` when one's set (TCP for telnet sessions) and falls back to UART when none is bound (boot, kernel logs, async tasks) — so the boot and test paths keep their UART output. **Telnet operators previously ran `model gpu` or `model bench` and saw nothing back**; this commit fixes that without disturbing the kernel log channel.

The existing partial workaround at `kernel/src/shell_sys.c:2425` (the `predicted: %d` `shell_printf` echo for `model infer-file`) becomes redundant once the underlying logits dump goes via `shell_printf`. Leaving it in place — it adds a useful summary line that the Rust path doesn't print; harmless duplication.

## Origin

Originally part of a single commit on the deploy branch. Split here so the runtime side can land independent of the GA10B-side `pipeline-mode/complete` gating that depends on the `GA10B_DBG` macro from #553.

## Test plan
- [x] `make test` — kprintf-float + shell-tcp suites still green; pre-existing `blob_autoload` failures unchanged from main.
- [x] Verified live on jetson-nano-2: `model gpu` and `model bench 1 1` both now show full output over telnet. Steady-state inference produces zero per-iteration log lines (engine + ga10b together once #553 also lands).

## See also

- #553 — companion PR; final piece of pipeline-mode/complete gating now lives there.

### DIFF
--- a/runtime/src/inference/engine.rs
+++ b/runtime/src/inference/engine.rs
@@ -808,10 +808,11 @@ pub unsafe fn run_inference(
     // any error so the user still gets an answer.
     let mut result: Result<usize, EngineError>;
     if mnist_gpu_fastpath_eligible(model_index) {
-        crate::log::log_info(b"[engine] mnist GPU fastpath eligible -- dispatching\0");
+        // Successful dispatch is the steady-state happy path; logging
+        // it every iteration drowns the console at >1 inf/s. The
+        // failure path is rare and operator-actionable, so it stays.
         match super::gpu::run_mnist_gpu_fastpath(input, input_len, output, output_len) {
             Ok(n) => {
-                crate::log::log_info(b"[engine] mnist GPU fastpath success\0");
                 let elapsed = kernel_ffi::get_time_ns().saturating_sub(start);
                 record_inference(elapsed);
                 return Ok(n);
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -96,6 +96,10 @@ fn rust_panic(info: &PanicInfo) -> ! {
         buf[..len].copy_from_slice(&file[..len]);
         buf[len] = 0;
 
+        // SAFETY: pure C variadic FFI calls. `fmt` is a static null-
+        // terminated byte literal whose `%s/%u/%u` specifiers match
+        // the (c-string, u32, u32) argument types. `buf` is a 256-byte
+        // null-terminated stack array that outlives both calls.
         unsafe {
             uart_printf(
                 b"at %s:%u:%u\n\0".as_ptr(),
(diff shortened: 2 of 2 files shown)


johnjezl:
Round-2 review response (commit f4e1825, on a rebased branch — pre-rebase HEAD was eabd5b4).

**Warning — panic handler dropped UART:** fixed. The panic handler now declares both `uart_printf` and `shell_printf` externs and emits the `at file:line:col` line to BOTH. UART first (most reliable, always-on, lab serial capture path), shell_printf second so the connected operator still sees it in their telnet session. `shell_printf` falls back to UART when no session is bound, so the double-emit is idempotent in that case.

**Suggestion — `print_logits_and_argmax` per-row drop hazard:** documented. Added a doc-ladder paragraph noting that each row is one `shell_printf` call (one ring enqueue), drop-oldest can elide rows under slow-drain conditions, and that for MNIST's 10 logits the trade-off is fine (~280 bytes, well under the 4 KB ring). Future callers with larger outputs (sched-MLP at 42 logits, N-class) should either batch or accept the partial-output trade-off explicitly. Comment lives at the helper so a new caller sees it.

**Deferred per the reviewer's 'out of scope' / 'follow-up' labels:**
- #567 — `mnist GPU fastpath failed` should be `log_warn`, not `log_info`
- #568 — hoist per-function `shell_printf` extern blocks to a single module-level extern
- #569 — `\\n` vs `\\r\\n` inconsistency sweep across runtime/

Branch was rebased onto current main to pick up the recent merges (#553, #555, #556, #565). The pre-rebase 12 blob_autoload failures were caused by the stale base; QEMU `make test` passes after the rebase + this commit.

I had to `--force-with-lease` to publish the rebased branch — first time on this PR, no upstream divergence.

PR #570, opened 2026-04-29
x86-64: fit model_mem pool to 256 MB QEMU RAM + add init smoke test

johnjezl:
## Summary

`rust_model_mem_init` requested 256 MB weights + 128 MB workspace on every target. On x86-64 QEMU q35 the guest only gets **256 MB total** (Makefile's `QEMU_MEMORY` for the X86_64 branch), so `pmm_alloc_pages(65536)` for the weight pool returned NULL, the Rust shim returned -1, and the model_mem allocator stayed uninitialized for the rest of the boot. Failure mode was effectively silent: `rust_model_alloc_weights` returned a null `ModelHandle`, callers that didn't check the handle silently no-op'd, and the only diagnostic was a single early-boot `[FAIL] Model memory init failed: PMM allocation failed` line.

This PR:

- **`runtime/src/lib.rs::rust_model_mem_init`** — gates pool sizes on `target_arch`. aarch64 keeps 256/128 MB; x86-64 gets 64/32 MB which fits the 256 MB budget with headroom for the kernel and Rust heap.
- **`kernel/tests/test_model_mem_smoke.c`** *(new)* — three-test suite that runs on every PLATFORM. Asserts both pools report `total_blocks > 0` post-init, and that a single weights+workspace alloc/free round-trip succeeds with 2 MB alignment. Wired in as the FIRST suite in `test_harness_run_all` so any future regression surfaces before downstream tests can silently degrade. Surface area is intentionally small — the existing `test_model_mem.c` (which doesn't run on x86-64 because some of its tests want 200 MB) covers the deeper invariants.
- **Docs** — `docs/model-memory.md` (per-target sizing table + rationale + checklist for raising sizes), `docs/specs/memory.md` (model_mem rows in the per-platform matrix), `docs/api/runtime.md` (FFI description points at the sizing section).

## Test plan

- [x] `make kernel PLATFORM=QEMU_VIRT` clean
- [x] `make kernel PLATFORM=RASPI5` clean
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO` clean
- [x] `make kernel PLATFORM=X86_64` clean
- [x] `make test` — new smoke suite passes (3/3) as the first test section
- [x] `make test PLATFORM=X86_64` — smoke suite passes (3/3) as the first test section, all the way through `test_workspace_alloc_round_trip`. Pre-existing flake in eviction-demo / linked_list_allocator tracked under #237 still fires later in the run; out of scope for this PR.

## What this does NOT fix

The remaining x86-64 `make test` flake — Rust panic in `linked_list_allocator-0.10.6/src/hole.rs:501` ("Freed node aliases existing hole! Bad free?") during `test_shell_cmd_eviction_demo`, ~80% repro rate — is heap-metadata corruption upstream of the eviction demo path. The model_mem sizing fix made it _reachable_ but didn't cause it. Investigation paths attempted (Vec lifecycle in `evict_and_retry`, `EvictedContentTracker::drain_expired`, trait-object dispatch through `with_active_policy`) didn't surface anything obviously wrong on inspection. Tracking under #237.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,10 @@ cmake-build-*/
 # Claude Code personal settings
 .claude/settings.local.json
 .claude/worktrees
+# Runtime artifacts (e.g. .claude/scheduled_tasks.lock) — never
+# checked-in; the only tracked content under .claude/ is settings,
+# commands, and notes, none of which use .lock.
+.claude/*.lock
 .codex
 
 # Codex local state
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -574,9 +574,10 @@ set(KERNEL_SOURCES
     # because the MOCK_CAMERA_FRAME option is defined further down.
     kernel/src/camera.c
     $<$<NOT:$<STREQUAL:${PLATFORM},X86_64>>:kernel/src/dtb.c>
-    # x86-64 has no DTB; provide minimal stubs for the two symbols
-    # platform-neutral code still references (dtb_get_memreserves /
-    # dtb_get_chosen).
+    # x86-64 stubs for dtb_get_memreserves / dtb_get_chosen — needed
+    # because kernel/mm/pmm.c and kernel/src/main.c call those
+    # symbols unconditionally, and kernel/src/dtb.c is excluded from
+    # the x86-64 build.
     $<$<STREQUAL:${PLATFORM},X86_64>:kernel/src/dtb_x86_stub.c>
     # General-purpose FDT reader — path + property lookup for any
     # driver. Built on every ARM64 target; x86-64 builds skip it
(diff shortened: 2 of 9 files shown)


johnjezl:
Addressed review feedback in `f0313c5c`:

- **S1** — `rust_model_mem_init` docstring reworded. The old text implied an uninitialized allocator caused the linked_list_allocator panic; the new text describes what actually happens (silent null-handle returns + silent no-ops in tests that don't check init), and points readers at `test_model_mem_smoke.c` rather than at #237.
- **S2** — Added an inline comment explaining the 64+32 MB choice (~150 MB headroom on x86-64 QEMU) plus the 'multiples of 2' invariant, so a future reader doesn't have to round-trip to docs to understand the literal.
- **S3** — `SmokeModelHandle` struct comment now explains why the duplication is deliberate (smoke runs first → layout drift fails here cleanly before the larger suite cascades), and names the refactor target if a third caller appears.

No functional changes; both ARM64 and x86-64 still build clean.

Issue #596, opened 2026-05-01
GA10B: residual within-run dispatch staleness on MNIST (PR #589 partial fix)


johnjezl:
## Symptom

PR #589 added `INVALIDATE_SHADER_CACHES` to the dispatch pushbuffer and substantially reduced GPU MNIST dispatch staleness, but **did not eliminate it**. Subsequent verification on a fresh kexec showed within-run mismatches reappearing at varying iteration positions (not exclusively iter-1).

## Updated evidence (jetson-nano-2, fresh kexec, 2026-05-01)

10× alternating fills (0.0f / 1.0f), expected steady-state: 1.0f → 5, 0.0f → 6:

| Run | iter 1 | iter 3 | iter 5 | iter 7 | iter 9 | mismatches |
|---|---|---|---|---|---|---|
| 1 | **6 ❌** | 5 | **6 ❌** | 5 | 5 | 2/5 |
| 2 | 5 ✓ | 5 | 5 | **6 ❌** | 5 | 1/5 |
| 3 | 5 ✓ | 5 | 5 | **8 ❌** | 5 | 1/5 |
| (with warmup) | **6 ❌** | 5 | 5 | **6 ❌** | 5 | 2/5 |

Notable:
- Mismatches occur at varying positions (1, 5, 7) — not just iter-1 as originally hypothesized.
- One run returned argmax=8, which is not a steady-state for either tested input. The wrong values are not strictly \"previous-input wins.\"
- A throwaway warmup dispatch in front of the loop did NOT eliminate the issue.

## Comparison

The original Phase 6 verification that motivated PR #589 showed iters 2-N as 100% deterministic across 4 runs. Either:
1. The prior verification got a lucky streak (sample size: 4 runs × 10 iters = 40 dispatches; not enough to expose a ~10-20% per-iter mismatch rate).
2. The fresh kexec / different DRAM placement of handoff buffers exposes an additional failure mode.
3. The SHADER_CACHES fix is partially effective (large reduction vs. pre-fix 4-5/10 mess, but not complete).

## What PR #589 did achieve

Pre-fix (per prior session summary): \"Run 1 OFF: 4/10 correct\" — 60% mismatch rate.
Post-fix (this session): ~10-20% mismatch rate concentrated on f=3F800000 input. Real improvement, not a no-op.

## Hypothesis space (updated)

The naïve \"previous-input-wins\" theory still fails. The argmax=8 case in particular suggests something more like:
[...]

PR #610, opened 2026-05-03
ga10b: always advance cached gp_put bookkeeping on poll failure (#596)

johnjezl:
## Summary

Band-aid fix for the multi-op MNIST pipeline wedge surfaced by re-measuring #596 at N=200. `ga10b_submit_and_poll` only advanced the cached `g_handoff.initial_gp_put` when `gp_advanced` was true. When the 2-second poll fired before PBDMA's GP_GET caught up, the cached put stayed un-advanced — but the GPFIFO entry and the USERD GP_PUT register had already been written. The next submit then computed the same slot, overwrote the in-flight entry, and re-wrote the same USERD value (a no-op for PBDMA). That's the "iter 1 op 8 fails, iter 2+ op 1 forever wedges" pattern from the [#596 update comment](https://github.com/SLM-OS/SLM-Operating-System/issues/596#issuecomment-4366724098).

Found via cross-reading nvgpu's submit path (`nvgpu_submit_prepare_gpfifo_track`, `gv11b_userd_gp_put`) — nvgpu always advances `c->gpfifo.put` after writing entries, regardless of the eventual fence outcome. Our cached counter has the same role and needs the same lifecycle.

## Empirical result on jetson-nano-2

Same alternating-fill probe (`/mnt/files/p.lua` at N=200, expected argmax 5 for 1.0f / 6 for 0.0f):

| Build | Outcome |
|---|---|
| Pre-fix (deployed `/root/slmos.elf` from 2026-04-30) | Wedge at iter 1 op 8; iter 2+ all wedge at op 1 → effectively t=200 / 200 mismatches |
| Post-fix (this PR's kernel) | Probe completes; **t=1 / 200 mismatches (0.5%)** |

A side-channel kernel panic in `net_pump` after the probe completed (PC alignment fault to an address whose bytes decode to Lua source from `/mnt/files/p.lua`) — likely a separate B1-class stack-overflow regression that PR #603 was supposed to close. Filing as a follow-up issue, not a blocker for this fix.

## Test plan

- [x] `make test-ga10b-bringup` (host) passes
- [x] `make test` (QEMU ARM64) passes
- [x] Built `make kernel PLATFORM=JETSON_ORIN_NANO`, deployed to `/root/slmos.elf`, kexec'd, ran probe — measured t=1/200
- [ ] Verify probe behavior is consistent across power-cycles (next session)
- [ ] Architectural follow-up: refactor `ga10b_dispatch_v7_pipeline` to nvgpu's bulk-submit pattern (separate PR)

## Why "band-aid"

This makes the wedge recoverable and unblocks measuring Bug C at scale, but doesn't change the per-op doorbell-and-poll pattern that's the architectural mismatch with how PBDMA is meant to be driven. The follow-up will bulk-queue all v7 pipeline op entries, write USERD GP_PUT once with the cumulative new value, ring the doorbell once, and poll only the final op's semaphore. That eliminates the per-op pressure on PBDMA and matches nvgpu's tested-at-scale pattern.

Refs #596.

### DIFF
--- a/kernel/gpu/nvidia/ga10b_bringup.c
+++ b/kernel/gpu/nvidia/ga10b_bringup.c
@@ -1517,13 +1517,24 @@ uint32_t ga10b_build_launch_kernel_with_sema_pushbuffer(uint32_t *pb,
  * be reported as success; an earlier loose criterion was a
  * now-resolved #273 workaround and has been dropped.
  *
- * GPFIFO bookkeeping: on any PBDMA progress (gp_advanced), update
- * `g_handoff.initial_gp_put` / `initial_gp_get` so a subsequent
- * submit targets a fresh slot. This is symmetric across success and
- * partial-failure paths because the slot has been consumed either
- * way; reusing it would overwrite a not-yet-drained entry only if
- * PBDMA hadn't seen our submit at all, and that case (!gp_advanced)
- * leaves the counters unchanged. */
+ * GPFIFO bookkeeping: ALWAYS advance `g_handoff.initial_gp_put` /
+ * `initial_gp_get` on return, regardless of `gp_advanced` or
+ * `payload_matched`. Once the GPFIFO entry has been written and
+ * USERD GP_PUT has been advanced, the slot is consumed from the
+ * channel's hardware perspective — a subsequent submit MUST target
+ * the next slot, or it would re-write the same USERD value (a
+ * no-op for PBDMA) and overwrite the in-flight entry. The earlier
+ * `if (gp_advanced)` guard caused a permanent wedge in #596 when
+ * the 2 s poll fired before PBDMA's GP_GET caught up. See the
+ * in-function comment for the full reasoning.
+ *
+ * Future architectural follow-up: refactor the v7 pipeline
+ * dispatcher to nvgpu's bulk pattern (queue all op entries, write
+ * USERD GP_PUT once with the cumulative new value, ring doorbell
+ * once, poll only the final semaphore). Eliminates per-op PBDMA
+ * pressure and matches the path NVIDIA tests at scale. The
+ * bookkeeping fix here remains load-bearing for any single-op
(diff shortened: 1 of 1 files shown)


johnjezl:
## Review actions addressed

1. **Stale function header comment** (Warning) — updated to describe the new always-advance behavior. Old text said `!gp_advanced` "leaves the counters unchanged" which was no longer true.
2. **Branch was forked from before #603 + #605** (Warning) — rebased onto `origin/main` (clean, no conflicts). New branch tip: `8ae64317`. `git merge-base --is-ancestor` confirms both #603 (54025f3e) and #605 (43e17a11) are now in this branch's history.
3. **Architectural follow-up note** (Suggestion) — added to the function header comment so future readers know this is a band-aid, not the long-term solution.
4. **Skipped:** stub host-test for the bookkeeping invariant. Low value given the change is one block; the empirical hardware re-verification below catches any regression more authoritatively than a mocked test would.

Force-pushed with `--force-with-lease`. Diff against main is unchanged in shape (one block in `ga10b_bringup.c`); only the header doc and inline comment grew.

## Hardware re-verification on rebased build

Built from `8ae64317`, deployed to jetson-nano-2, kexec'd. Same alternating-fill probe at N=200 in lua-admin.

**Result: t=1/200 (same 0.5% mismatch rate as the pre-rebase run). NO net_pump panic. All task canaries intact post-probe.** Full transcript [posted to #596](https://github.com/SLM-OS/SLM-Operating-System/issues/596#issuecomment-4367202277).

Three things confirmed simultaneously:

| | |
|---|---|
| Band-aid works at GPU layer | Probe runs to completion; wedge gone |
| #603 closed Bug B1 end-to-end | net_pump panic from previous run does NOT recur with the heap fix in |
| #605 canary check is healthy | Post-probe `canary` shell cmd shows all 8 task stacks intact |

Ready for merge once /slmos-review confirms the rebase + comment updates.

Issue #601, opened 2026-05-02
GA10B: GPU pipeline wedge after ~50 cumulative compute dispatches


johnjezl:
> **Note (2026-05-02):** Originally filed as a hard wedge that panicked the system. The panic half turned out to be a separate bug — a stack overflow in `lua_slm_dofile` corrupting net_pump's stack — and was fixed in PR #603. This issue now tracks ONLY the remaining GPU-side wedge.

## Symptom

After ~50-60 cumulative compute dispatches across multiple `lua-admin` invocations on the same kexec session, the GPU pipeline wedges. The first failing op reports:

```
[GA10B-P8] GP_GET advanced but payload didn't land (got 0x0, want 0xcafedead) —
PBDMA consumed our submit but the GPU didn't produce the expected side effect.
[GA10B-P8] pipeline op 1 failed (rc=-1) — aborting chain
```

After PR #603, this no longer panics the system — the shell prompt remains usable, canaries on all task stacks are intact. But every subsequent dispatch fails until we re-kexec.

## Reproduction (post-PR-#603)

```bash
slmos-kexec --no-gpu-suspend slmos.elf
# stage alternating-fill probe (10 dispatches per invocation)
lua-admin /mnt/files/p.lua  # run #1: 10 OK
lua-admin /mnt/files/p.lua  # run #2: 10 OK
lua-admin /mnt/files/p.lua  # run #3: 10 OK
lua-admin /mnt/files/p.lua  # run #4: 10 OK
lua-admin /mnt/files/p.lua  # run #5: 10 OK
lua-admin /mnt/files/p.lua  # run #6: ~3 dispatches OK, then wedge mid-iteration
```

GP_GET advances → PBDMA processed the entire pushbuffer.
Semaphore payload didn't land → either the dispatch never ran, or it ran but didn't write the semaphore.

## What's known

- **Cumulative across `lua-admin` invocations** — channel state persists across them; the wedge is in GPU state, not SLM-OS state
- **PR #589** (per-launch INVALIDATE_SHADER_CACHES) extended the safe budget from ~70 dispatches (pre-fix) to ~50 dispatches (post-fix) — partial mitigation, not a fix
- **System remains alive** post-wedge after PR #603 — only the GPU is unusable, not the kernel
- Pool slot wraparound is NOT the cause: 1024 slots, we wedge at ~400
[...]

Issue #580, opened 2026-04-30
gpu(ga10b): make ga10b_dispatch_v7_pipeline directly host-testable


johnjezl:
## Background

`ga10b_dispatch_v7_pipeline` (`kernel/gpu/nvidia/ga10b_bringup.c:1755`) was extracted from `ga10b_bringup_launch_kernel` in PR #574 to cut function size and improve readability. The docstring acknowledges that direct host-testability of the helper is a follow-up:

> Extracted from the launch_kernel function body for readability + (eventually) testability — the bounds-check error paths have dedicated host-test coverage via `ga10b_v7_validate_handoff`, which this function delegates to.

The helper currently reads `g_handoff` and `g_qmd_pool_next_slot` directly from globals, and calls `gsp_platform->cache_invalidate` / `cache_clean` / `mb()` plus `ga10b_submit_and_poll` — none of which are easily mocked from a host test without simulating the whole MMIO surface.

The validator path (`ga10b_v7_validate_handoff`) is host-tested end-to-end (10 cases). The dispatch loop body is covered indirectly via `ga10b_qmd_pool_prepare`'s tests (slot rotation, byte-equivalence to direct populate, etc.). What's not directly tested is the integration: "given a v7 handoff with N ops, does the helper iterate correctly, advance the slot counter, and submit each op exactly once with the right pushbuffer arguments."

## Proposed approach

Refactor `ga10b_dispatch_v7_pipeline` to take its inputs as parameters instead of reading globals:

```c
static int ga10b_dispatch_v7_pipeline(
    struct ga10b_bringup *b,
    const struct ga10b_channel_handoff *handoff,
    uint32_t *next_slot_inout,
    int (*submit_fn)(struct ga10b_bringup *b, const uint32_t *pb_buf,
                     uint32_t pb_dwords, uint64_t poll_phys,
                     uint32_t expected_payload, int error_phase,
                     const char *tag));
```

Production caller passes `&g_handoff`, `&g_qmd_pool_next_slot`, and `ga10b_submit_and_poll`. Host test passes a synthetic handoff, a local slot counter, and a stub `submit_fn` that records its arguments into a test-owned ring buffer.

[...]

PR #612, opened 2026-05-03
ga10b: refactor v7 dispatcher to nvgpu-style bulk submit (#596)

### LINKED ISSUE #596: GA10B: residual within-run dispatch staleness on MNIST (PR #589 partial fix)

johnjezl:
## Summary

Follow-up to PR #610's band-aid. Replaces the per-op submit-and-poll loop in `ga10b_dispatch_v7_pipeline` with the nvgpu-style bulk submit pattern: queue all N kernel-dispatch pushbuffers + a trailing `COMPUTE_B REPORT_SEMAPHORE` release into the GPFIFO, advance USERD GP_PUT once to the cumulative new value, ring the doorbell once, poll the trailing sema for chain completion.

Reference: `nvgpu-common-fifo-submit.c::nvgpu_submit_prepare_gpfifo_track` (lines 392-413, 572) + `nvgpu-hal-fifo-userd_gv11b.c::gv11b_userd_gp_put` (lines 67-77) in the local `slmos-reference-cache`.

## Empirical result

Same alternating-fill probe at N=200 on jetson-nano-2:

| Build | Mismatches | Total runtime |
|---|---|---|
| pre-fix (deployed 2026-04-30) | 200 / 200 | wedged |
| per-op + band-aid (PR #610) | 1 / 200 | ~360 sec |
| **bulk submit (this PR)** | **0 / 200** | **65 ms** |

Two findings:

1. **Bug C is closed end-to-end.** The 0.5% residual mismatch from PR #610 was caused by the per-op pattern itself — batching into a single submit eliminates both the wedge AND the residual. Hypothesis 2 from #596 (per-op timing/cache interactions) is confirmed; Hypothesis 1 (separate L2 invalidate) is no longer needed.
2. **~5500× speedup as a side benefit.** Per-op spent ~2 sec in poll timeouts on flaky iterations + 8 doorbell+poll cycles per inference. Bulk does one of each per chain, drains in <1 ms per inference.

Diagnostic post-probe: all 8 task stack canaries intact.

## What changed

- New layout: `pushbuf_phys` packs N × `GA10B_LAUNCH_KERNEL_PB_DWORDS` (14 dw / 56 B each) followed by 1 × `GA10B_COMPUTE_SEMA_RELEASE_PB_DWORDS` (12 dw / 48 B).
- Capacity check: refuses chains where `N+1 > gpfifo_entries/2` (half-ring safety budget mirroring nvgpu's `EXTRA_GPFIFO_ENTRIES` margin) or where total pushbuffer bytes exceed `pushbuf_size`. With the post-#601 ring (512 entries) and the helper's 65 KB pushbuffer, allows up to 255 ops per chain.
- Cache-clean handles ring-wrap (split into two contiguous runs if the entries cross the boundary).
- Bookkeeping: matches PR #610's always-advance contract — `g_handoff.initial_gp_put` and `initial_gp_get` advance after the chain regardless of poll outcome.
- The trailing release uses `NVC7C0_REPORT_SEMAPHORE` with `STRUCTURE_SIZE_ONE_WORD` + `FLUSH_DISABLE=0` (Mesa NVK `nvk_dispatch_signal_semaphore` convention) — drains the compute pipeline AND flushes L2→DRAM before release. So the single sema poll is the unambiguous "all N kernels done, all outputs in DRAM" signal.

## What didn't change

- Single-op callers (Phase 7 sema/compute smoke tests) still use `ga10b_submit_and_poll` directly. That path keeps PR #610's always-advance bookkeeping for any future single-op caller.
- The pure-logic builders (`ga10b_build_launch_kernel_pushbuffer`, `ga10b_build_compute_sema_release_pushbuffer`) are unchanged — the bulk path uses both, the per-op path used the with-sema variant.
- Existing host tests still cover all the building blocks. The bulk dispatch itself is hard to host-test without the refactor in #580 (parameterize the handoff + slot counter + stub-submit fn pointer); empirical hardware verification covers it for now.

## Failure mode tradeoff

Per-op error localization is lost — a poll timeout means "somewhere in the chain stalled". The diagnostic on timeout still discriminates two cases:
- `GP_GET unchanged` → PBDMA didn't see our submits at all (channel state issue)
- `GP_GET advanced but sema didn't fire` → one of the kernels faulted internally

For deeper localization, re-run with `gpu debug on` to see per-op QMD/pb authoring traces.

## Test plan

- [x] `make test-ga10b-bringup` (host) passes
- [x] `make test` (QEMU ARM64) passes
- [x] Built `make kernel PLATFORM=JETSON_ORIN_NANO`, deployed to `/root/slmos.elf`, kexec'd
- [x] Ran 200-iter alternating-fill probe — `m=0 ms=65`
- [x] Post-probe `canary` shell command — all 8 task stacks intact
- [ ] Long-running stability run (post-merge, equivalent of the 216M-iteration probe from #607's verification)

## References

- Closes #596 (Bug C residual within-run staleness on MNIST)
- Builds on PR #610 (band-aid bookkeeping fix; single-op path retains it)
- nvgpu reference: `nvgpu-common-fifo-submit.c::nvgpu_submit_prepare_gpfifo_track` + `nvgpu-hal-fifo-userd_gv11b.c::gv11b_userd_gp_put`
- Mesa reference: `nvk_dispatch_signal_semaphore` (FLUSH_DISABLE=0 + STRUCTURE_SIZE_ONE_WORD)

### DIFF
--- a/kernel/gpu/nvidia/ga10b_bringup.c
+++ b/kernel/gpu/nvidia/ga10b_bringup.c
@@ -51,6 +51,28 @@ bool ga10b_dispatch_verbose_get(void)
             uart_printf(__VA_ARGS__);                               \
     } while (0)
 
+/* Dispatch poll budget shared by ga10b_submit_and_poll (single-op
+ * smoke-test path) and the v7 bulk dispatcher. Each iteration of
+ * the outer poll loop pairs with one ga10b_dispatch_poll_delay_us
+ * call below, so the wall-clock cap is approximately
+ * GA10B_DISPATCH_POLL_TIMEOUT_US microseconds. The 2 s value is
+ * generous — typical sema release fires in tens of microseconds,
+ * but tail latencies under context-switch pressure or first-op
+ * QMD fault diagnostics can stretch into the hundreds of ms. */
+#define GA10B_DISPATCH_POLL_TIMEOUT_US 2000000u
+
+/* Approximate 1 µs delay via volatile nop-spin. Calibrated for
+ * Cortex-A78AE (Jetson Orin); not precise wall-clock — the actual
+ * wait depends on CPU clock + memory pressure — but bounded above
+ * by a small multiple of 1 µs so the overall poll budget tracks
+ * GA10B_DISPATCH_POLL_TIMEOUT_US within a constant factor. The
+ * `volatile` prevents the compiler from optimizing the loop body
+ * away. */
+static inline void ga10b_dispatch_poll_delay_us(void)
+{
+    for (volatile int i = 0; i < 1500; i++) { }
+}
+
 extern const struct gsp_platform_ops *gsp_platform;
 
 /* ---- Firmware symbols from nvidia_ga10b_firmware.S ----
@@ -1660,7 +1682,7 @@ static int ga10b_submit_and_poll(struct ga10b_bringup *b,
 
     GA10B_DBG("[%s] polling (2s timeout)...\n", tag);
     uint32_t poll_val = 0;
-    for (uint32_t us = 0; us < 2000000; us++) {
+    for (uint32_t us = 0; us < GA10B_DISPATCH_POLL_TIMEOUT_US; us++) {
         if (gsp_platform->cache_invalidate) {
             gsp_platform->cache_invalidate((void *)poll, sizeof(uint32_t));
         }
@@ -1669,7 +1691,7 @@ static int ga10b_submit_and_poll(struct ga10b_bringup *b,
          * tests can pin the exact predicate without re-encoding it
          * (and the Linux launcher uses the same helper). */
         if (ga10b_poll_match(poll_val, expected_payload)) break;
-        for (volatile int i = 0; i < 1500; i++) { }
+        ga10b_dispatch_poll_delay_us();
     }
 
     uint32_t gp_get_word = g_handoff.userd_gp_get_offset / 4;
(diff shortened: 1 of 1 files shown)


johnjezl:
## Review polish addressed

**Suggestions 2 + 3** (magic timeout / busy-wait deduplication): introduced two file-local entries near the top of `ga10b_bringup.c`:

- `#define GA10B_DISPATCH_POLL_TIMEOUT_US 2000000u` — replaces the literal `2000000` at both poll-loop call sites (`ga10b_submit_and_poll` and the v7 bulk dispatcher).
- `static inline void ga10b_dispatch_poll_delay_us(void)` — encapsulates the `for (volatile int i = 0; i < 1500; i++)` Cortex-A78AE-calibrated ~1 µs nop-spin. Replaces the inline copy at both call sites.

Both helpers carry doc-comments on the calibration assumption and the volatile rationale. Other 1500-iter busy-waits in the file (ACR sequence, FECS, GPCCS bring-up — lines 506, 524, 541, 761, 978) are in different code paths from a different era; not in scope for this PR.

**Suggestion 1** (function size): per the review's own "Acceptable as-is" note, no change. The header docstring breaks the function into 6 explicit phases and most state flows through every phase — splitting into helpers would just push state-passing into multi-arg signatures.

`make test-ga10b-bringup` and `make test` both green after the amend. Hardware run signal from the prior verification still applies — same kernel + dispatcher logic, just deduplicated literals.

Force-pushed with `--force-with-lease`. New tip: `358c74ec`.

Issue #475, opened 2026-04-27
SLM integration on Jetson Orin Nano — Qwen2.5-1.5B end-to-end (umbrella)


johnjezl:
## Overview

End-to-end support for Qwen2.5-1.5B-Instruct on Jetson Orin Nano Super: GGUF upload via VFS → CLI launch → prompt → streaming response → live perf monitoring → GPU-accelerated inference with Ampere HMMA tensor cores.

## Specs

- Plan: `docs/plans/slm-integration-plan.md`
- Spec: `docs/specs/slm-integration.md`
- GPU handoff design: `docs/specs/gpu-slm-handoff.md`
- User tutorial: `docs/tutorials/slm-prompt.md`

## Status (2026-04-27)

All ten milestones structurally complete on the SLM feature branch.

| Milestone | Status                              | PR    |
|-----------|-------------------------------------|-------|
| M0 — Host tooling & model provisioning | ✅ Shipped | #492  |
| M1 — GGUF v3 parser                     | ✅ Shipped | #497  |
| M2 — BBPE tokenizer (Qwen)              | ✅ Shipped | #502  |
| M3 — Q4 quantization kernels            | ✅ Shipped | #505  |
| M4 — Transformer operators (CPU)        | ✅ Shipped | #520  |
| M5 — Decoder, KV cache, sampler, session | ✅ Shipped | #524  |
| M6 — GPU acceleration scaffolding       | ✅ Scaffolding shipped | #526  |
| M7 — Shell integration & streaming I/O  | ✅ Shipped | #528  |
| M8 — slm-runner component               | ✅ Shipped | #530  |
| M9 — Telemetry, benchmarks, polish      | ✅ In flight (this PR) | TBD |

## Demo readiness

End-to-end real text generation lands when **M5.3** + the Jetson hardware deploy converge. Today's stub `forward_step` exercises the full plumbing (load → launch → tokenize → KV cache append → stream → telemetry) but emits zero-logit greedy tokens.

## Deferred work — follow-up issues

- **#538** — M5.3: real `forward_step` (registry tensor data + per-session tokenizer + weight pool through M4 op chain). **P1, blocks demo.**
- **#539** — M6.A-2..6: pre-kexec L4T loader + bare-metal pushbuffer dispatch + smoke-test gate + cache coherency + PTIMER. P2.
[...]

PR #620, opened 2026-05-04
slm_ffi: kind-parameterized GPU dispatch FFI

johnjezl:
## Summary

The kernel-side \`ga10b_*\` dispatch API has been kind-parameterized since \`pipeline_kind\` landed (#475 / PR-3 of \`gpu-policy-models.md\`), but \`slm_ffi.c\` had two near-duplicate copies of the dispatch plumbing — one for MNIST and one for sched-MLP. Each new model was paying a copy-paste-modify tax: a \`g_<model>_bringup\` static, an \`ensure_<model>_bringup\` helper, plus the \`slm_gpu_run_<model>\` / \`slm_gpu_set_<model>_input\` entry points. Adding a third model (e.g. Qwen for #475) would have meant copying it again.

This consolidates the duplicated bodies into one kind-indexed array + one kind-parameterized API, with the existing MNIST / sched-MLP named entry points retained as one-line shims for backward compatibility.

## What's new

In \`kernel/include/slm_ffi.h\` (and stubbed in non-Jetson builds):

\`\`\`c
int slm_gpu_run(uint32_t kind, void *output, size_t cap);
int slm_gpu_set_input(uint32_t kind, const void *bytes, size_t cap);
int slm_gpu_set_input_fill(uint32_t kind, uint32_t bits, uint32_t n);
int slm_gpu_run_with_input(uint32_t kind,
                            const void *input, size_t input_cap,
                            void *output, size_t output_cap);
\`\`\`

\`kind\` is one of \`enum ga10b_pipeline_kind\` (\`GA10B_PIPELINE_KIND_MNIST\`, \`_SCHED_MLP\`, \`_EVICTION_QNET\`, ...). Caller picks the pipeline; SLM-OS routes via the kind discriminator stamped into the handoff by the host-side helpers.

## What's gone

- \`g_mnist_bringup\` / \`g_sched_bringup\` file-scope statics → \`g_bringups[GA10B_PIPELINE_KIND_COUNT]\` indexed by kind.
- \`ensure_mnist_bringup\` / \`ensure_sched_bringup\` near-duplicate bodies → single \`ensure_bringup(uint32_t kind)\`.
- \`slm_gpu_run_mnist\` / \`slm_gpu_set_mnist_input\` / \`slm_gpu_set_mnist_input_fill\` / \`slm_gpu_run_sched_inference\` dispatch-loop bodies → 1-line forwards to the generic functions.

## What stays

- The kind-named shims (\`slm_gpu_run_mnist\` etc.) keep their existing signatures + symbols, so Lua bindings (\`l_gpu_run_mnist\`), Rust FFI (\`kernel_ffi::gpu_run_mnist\`), and the engine.rs MNIST fastpath all keep working unchanged.
- \`g_gpu_dispatch_lock\` is still held across the full set_input + launch + read sequence for kinds that need atomicity — \`slm_gpu_run_with_input\` matches the previous combined \`run_sched_inference\` pattern.
- The Bug A warmup is still per-kind (each cold-start needs one throwaway dispatch), now living in the single \`ensure_bringup\` body rather than duplicated.
- Circuit-breaker discipline is unchanged — every entry point routes through \`gpu_dispatch_record_result\` / \`gpu_dispatch_record_set_input_result\` the same way.

## Enum addition

Added \`GA10B_PIPELINE_KIND_COUNT\` as a sentinel at the end of \`enum ga10b_pipeline_kind\`. Used solely for sizing kind-indexed arrays in SLM-OS; never written to a wire handoff. Existing values (\`MNIST=0\`, \`SCHED_MLP=1\`, \`EVICTION_QNET=2\`) unchanged. Adding a new kind: insert above \`COUNT\`, the array sizes itself.

## Net delta

\`\`\`
3 files changed, 195 insertions(+), 119 deletions(-)
\`\`\`

~180 lines of duplication removed, ~80 lines of generic API + thin shims added.

## Test plan

- [x] \`make test-ga10b-bringup\` (host) passes
- [x] \`make test\` (QEMU ARM64) passes
- [x] \`make kernel PLATFORM=JETSON_ORIN_NANO\` builds clean
- [ ] Hardware verification: re-run alternating-fill probe at N=200 — predict identical \`m=0 ms=65\` to PR #612 since the dispatch behavior is byte-identical, only the dispatch shim is consolidated.

## What this unblocks

Adding GPU support for a new model (Qwen, future SLMs) is now:
1. Add a \`GA10B_PIPELINE_KIND_<NEW>\` enum value
2. Have the host helper write a handoff with that kind
3. Call \`slm_gpu_run(NEW_KIND, ...)\` from the runtime

No new file-scope statics, no new \`ensure_*_bringup\` body, no new \`slm_gpu_run_*\` entry points.

### DIFF
--- a/kernel/gpu/nvidia/ga10b_channel_handoff.h
+++ b/kernel/gpu/nvidia/ga10b_channel_handoff.h
@@ -239,6 +239,12 @@ enum ga10b_pipeline_kind {
     GA10B_PIPELINE_KIND_MNIST         = 0,
     GA10B_PIPELINE_KIND_SCHED_MLP     = 1,
     GA10B_PIPELINE_KIND_EVICTION_QNET = 2,
+    /* Sentinel — number of currently-defined kinds. NOT a valid
+     * value for the wire `pipeline_kind` field (handoffs only ever
+     * carry an actual kind). Used by SLM-OS for sizing kind-indexed
+     * arrays such as the per-kind bringup pool in slm_ffi.c. Adding
+     * a new kind: add it above this line, COUNT picks up automatically. */
+    GA10B_PIPELINE_KIND_COUNT
 };
 
 /* One entry per op in a v5/v6 pipeline. SLM-OS reads this array from
--- a/kernel/include/slm_ffi.h
+++ b/kernel/include/slm_ffi.h
@@ -826,6 +826,32 @@ int slm_gpu_get_info(RustGpuInfo *info);
  */
 int slm_gpu_run_mnist(void *logits_bytes_out);
 
+/* ---- Generic kind-parameterized GPU dispatch ----
+ *
+ * Each function takes a `kind` selector matching one of
+ * `enum ga10b_pipeline_kind` (GA10B_PIPELINE_KIND_MNIST,
+ * GA10B_PIPELINE_KIND_SCHED_MLP, GA10B_PIPELINE_KIND_EVICTION_QNET,
+ * ...). The caller chooses which pre-uploaded pipeline to dispatch;
+ * SLM-OS routes to the matching handoff via the kind discriminator
+ * the helpers stamp into DRAM.
+ *
+ * The kind-named entry points above (slm_gpu_run_mnist,
+ * slm_gpu_run_sched_inference) are 1-line shims for backward
+ * compatibility — new model integrations should call these generic
+ * functions with the appropriate kind constant.
+ *
+ * Output / input caps are caller-supplied: callers know the model
+ * shape, the kernel just bounds-checks against the handoff's
+ * input_buf_size (set_input) or output_size (run). On non-Jetson
+ * platforms all four return -1 unconditionally. */
+int slm_gpu_run(uint32_t kind, void *output, size_t output_cap);
+int slm_gpu_set_input(uint32_t kind, const void *bytes, size_t cap);
+int slm_gpu_set_input_fill(uint32_t kind, uint32_t value_bits,
+                           uint32_t n_floats);
+int slm_gpu_run_with_input(uint32_t kind,
+                           const void *input, size_t input_cap,
+                           void *output, size_t output_cap);
+
 /*
  * FP-free argmax over an array of fp32 bit patterns. Used by
  * Lua / shell callers that need the predicted class but can't do
--- a/kernel/src/slm_ffi.c
+++ b/kernel/src/slm_ffi.c
@@ -355,28 +355,30 @@ int slm_gpu_get_info(RustGpuInfo *info)
  */
 
 #ifdef PLATFORM_JETSON_ORIN_NANO
-/* Persistent bringup state for repeat MNIST + sched MLP runs. First
- * call walks inherit + channel; subsequent calls reuse the
- * channel-open state and only re-dispatch the kernels.
+/* Per-kind dispatch lock — serialises every `slm_gpu_*` entry
+ * point (run / set_input / set_input_fill / run_with_input and
+ * their kind-named shims) so only one dispatch is in flight at a
+ * time, regardless of which CPU the caller runs on.
  *
- * Concurrency: PR-3 of gpu-policy-models.md widened the caller
- * surface for the sched path to `ai_mlp_assign_cpu`, which the
- * scheduler invokes from any CPU during `scheduler_add_task` (no
- * scheduler-wide lock). Both `slm_gpu_run_mnist` and
- * `slm_gpu_run_sched_inference` mutate the file-scope `g_handoff`
- * (in ga10b_bringup.c) and their per-kind bringup globals. Without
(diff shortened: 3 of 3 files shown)


johnjezl:
## Round-2 review actions addressed + hardware verification finding

### Suggestions addressed (in the amended commit \`1dea2f3f\`)

- **Stale comment block** above \`g_gpu_dispatch_lock\`: rewritten to reference the generic \`slm_gpu_*\` entry points instead of the specific MNIST/sched-named ones. The lock's role is unchanged; only the wording.
- **Silent invalid kind**: \`ensure_bringup\` now \`uart_printf\`s when called with \`kind >= GA10B_PIPELINE_KIND_COUNT\` so a buggy caller (off-by-one, uninitialized variable) is visible in logs instead of returning a silent -1.

Skipped the hardcoded-output-caps suggestion per the review's "acceptable as-is" note.

### Hardware verification — REGRESSION CONFIRMED

Built kernel from this branch alone (no dispatcher), kexec'd to jetson-nano-2, ran the 10-iter alternating-fill probe via TCPSH:

\`\`\`
iter 1 input=0 argmax=3 expected=6   ← WRONG
iter 2 input=1 argmax=5 expected=5
iter 3 input=0 argmax=6 expected=6
iter 4 input=1 argmax=5 expected=5
... iter 5..10 all correct
DONE m=1/10
\`\`\`

So the iter-1 anomaly I saw earlier on the dispatcher branch (\`feat/gpu-dispatcher-task\`) is **inherited from this cleanup**, not introduced by Option 2. PR #612's verification (per its merge notes) showed \`m=0/200\` on the same probe shape — so this is a real regression introduced by the cleanup.

### What's different vs PR #612 that could matter

The cleanup's only architectural change is \`g_mnist_bringup\` / \`g_sched_bringup\` (separate file-scope statics) → \`g_bringups[GA10B_PIPELINE_KIND_COUNT]\` (one indexed array). Function bodies are otherwise byte-identical. The bug is most likely:

1. **BSS layout shift** affecting cache-line aliasing with \`g_handoff\` or another dispatch-relevant global, OR
2. **A pre-existing iter-1 race** that was masked by the old layout's timing and is now exposed.

Iter 1 produces a different "wrong" value each time I've measured it (\`argmax=5\` on the dispatcher branch run, \`argmax=3\` on this run), which is consistent with a stale-state read whose source depends on timing rather than a deterministic field-write bug.

### Recommendation

Hold this PR. The iter-1 mystery needs root-causing before merge — either by reproducing reliably with diagnostics or by reverting the layout change (single statics) and confirming iter 1 is back to correct. If the BSS-layout theory holds, a fix could be as small as adding alignment padding to \`g_bringups\`; if it's a latent race the fix is harder.

The doc/diagnostic improvements are independent and don't depend on the iter-1 outcome — could land separately if there's appetite to split. But since the structural cleanup is the headline, simpler to fix the regression here than to fork it.

Branch \`refactor/gpu-ffi-kind-generic\` preserved at \`1dea2f3f\` for the iter-1 investigation work.

johnjezl:
## Correction — cleanup is INNOCENT, iter-1 issue is in \`origin/main\`

I retract the merge-block recommendation from the previous comment. The iter-1 issue is **not** introduced by this PR — it's already present in \`origin/main\`.

### Cross-check on hardware

Built a kernel from \`origin/main\` (no cleanup applied) and ran the same 10-iter alternating-fill probe via TCPSH on jetson-nano-2:

| Trial | Result |
|---|---|
| Run 1 (lua-admin invocation 1) | \`m=1/10\` (iter 1 wrong) |
| Run 2 (same kernel, fresh TCPSH) | \`m=1/10\` |
| Run 3 (same kernel, fresh TCPSH) | \`m=1/10\` |
| (power cycle + kexec) | |
| Run 4 | \`m=0/10\` (iter 1 correct) |
| Run 5 | \`m=1/10\` |
| Run 6 | \`m=1/10\` |

5 of 6 runs show iter 1 wrong. So:
- It's already in main.
- It's intermittent — not deterministic.
- Failure rate is ~83% at this sample size.

PR #612's "m=0/200" headline was a single run that got lucky (probability ≈ 17% per the above). The bug was always there; the larger N just hid it because mismatches are concentrated on iter 1, which is "1 in 200" → exact same m=1 we'd get on the unlucky run, but framed against 200 means rounding error.

Closing #596 was premature — the residual iter-1 mismatch is still real. Will reopen / comment with the new evidence after this.

### Implication for PR #620

Strictly a cleanup, no behavior change, no regression introduced. The two earlier round-2 suggestions are addressed (stale comment block + invalid-kind diagnostic). Test plan steps 1-3 pass; the hardware step now shows the cleanup behaves identically to main on this probe (both flake on iter 1).

**Withdrawing the merge-block.** PR #620 is ready as-is.

The iter-1 race is its own thing — separate issue, separate investigation, doesn't gate this cleanup.

Issue #635, opened 2026-05-06
Pi 5 / Jetson: extend COOP_PREEMPT to cover pure CPU-bound loops via slm_preempt_point()


johnjezl:
## Context

`COOP_PREEMPT` (PR #126, issue #99) drives the scheduler's tick from `schedule()` entry: a CNTPCT-based check synthesizes a `scheduler_tick()` call every 10 ms. This works for any workload that yields — message-router consumers, lock-using tasks, sleep callers — but a pure CPU-bound loop with no yield monopolizes its CPU. That gap is the documented "limitation" in `docs/archive/investigations/pi5-preemption-resolution.md` §"What this does NOT unlock".

Issue #134 owns the long-term path back to hardware timer IRQ delivery (custom EL3 armstub). This issue is the cheap, kernel-side complement: a one-call cooperation point that any in-tree CPU-bound loop can invoke at its back-edge to let the scheduler intervene.

## Approach

`slm_preempt_point()` macro in `kernel/include/preempt_point.h`:

- **`COOP_PREEMPT` platforms (Pi 5, Jetson):** expands to `slm_preempt_check_and_yield()`. Reads `CNTPCT_EL0`, compares against the per-CPU 10 ms deadline. Cost on the not-due path: ~5–8 instructions, branch not taken. On the due path: full `schedule()` invocation.
- **Non-`COOP_PREEMPT` platforms (QEMU, x86-64):** expands to `((void)0)`. Call sites remain unconditional.

Diagnostic counters `slm_preempt_point_calls` and `slm_preempt_point_schedule_calls` track fast/slow path entries for tests and runtime diagnostics.

## Acceptance criteria

- [x] Macro defined in `kernel/include/preempt_point.h`
- [x] Helper in `kernel/sched/sched.c` next to `coop_preempt_maybe_tick`
- [x] Call sites added to `kernel/tests/test_integration.c:delay()`, `kernel/ai_accel/hailo/hailo_control.c:wait_for_response`, `kernel/ai_accel/hailo/hailo_vdma.c` poll loops
- [x] Soak test `test_slm_preempt_point_drives_schedule` in `kernel/tests/test_coop_preempt.c` (COOP_PREEMPT-only): 50 ms tight loop of macro calls must produce ≥ 4 schedule() entries
- [x] Smoke test `test_slm_preempt_point_callable` (all platforms): 1 024 invocations must not fault
- [x] Policy documented in `docs/scheduler.md` §"Preemption Model"
[...]

Issue #134, opened 2026-04-13
Pi 5: restore hardware timer IRQ delivery (armstub / ICC_SRE investigation)


johnjezl:
## Context
#99 resolved by \`PI5_COOP_PREEMPT\` — cooperative preemption via \`CNTPCT_EL0\` at yield points. That meets capstone requirements but leaves a known limitation: pure CPU-bound tasks with no yield points monopolize their CPU.

Restoring true hardware-preempted scheduling on Pi 5 would require debugging the Pi 5 firmware / TF-A interaction that currently prevents timer IRQ delivery to EL1. The investigation in \`docs/pi5-preemption-resolution.md\` eliminated every approach reachable from EL1/EL2:

- \`GICD_IGROUPR\` post-write readback reads as 0 from non-secure (spec-mandated, not evidence of silent discards — timer IRQ really is Group 1 per \`HPPIR=30\`).
- \`GICC_CTLR\` bypass-disable bits correctly set; not the issue.
- \`ICC_SRE_EL2\` access from EL2 traps to EL3 (\`ICC_SRE_EL3.Enable=0\`) and TF-A halts the CPU.
- \`ICC_SRE_EL3\` writes UNDEF on this BCM2712 Cortex-A76 configuration.
- Re-enabling \`armstub8-2712.bin\` (even minimal pass-through) breaks PSCI — boot hangs at \"CPU map: 4 CPUs configured\" before secondary bring-up. The ~60% boot-garble rate documented when the armstub was originally disabled remains unsolved.

## Work
1. Reverse-engineer Pi 5 firmware + TF-A armstub hand-off ABI. Identify why PSCI SMCs fail when an armstub is present.
2. Fix the armstub boot-garble (required either way).
3. Once the armstub is functional, configure \`GICD_CTLR\` / \`ICC_SRE_EL3\` / \`SCR_EL3\` at EL3 such that timer IRQ 30 delivers to EL1.
4. Flip \`PI5_COOP_PREEMPT\` default to OFF once hardware IRQs are reliable.

## Benefit
- True mid-execution preemption of CPU-bound tasks.
- The merged ELR trampoline (PR #98) activates automatically and gets exercised under load.
- Removes the yield-pointiness requirement from workload code.

## Non-blocker
- Cooperative preemption is sufficient for the capstone.
- Tasks that use the message router, locks, sleep, UART, or any standard OS primitive naturally yield and are preempted adequately under coop-preempt.
[...]

Issue #99, opened 2026-04-13
Pi 5: Timer IRQs (and all exceptions) not delivered despite DAIF.I=0


johnjezl:
## Problem

On Pi 5, no exception of any kind fires at runtime even when tasks run with \`DAIF.I=0\` and the GIC shows the timer IRQ pending. \`timer_handler_count\` stays at 0 indefinitely, and the kernel achieves preemption only through cooperative yields and idle WFE/SEV.

This is a pre-existing issue surfaced during #57 (secondary-CPU preemption) investigation but independent of that work. The previously documented fix in commit \`ac46e40\` (unmask IRQs in \`task_entry_trampoline\` + remove boot-stack \`daifclr\`) does not resolve it — it was noted as \"causes boot hang\" and remained unapplied. This blocker explains why.

## Evidence

- \`cpu\` shell output on Pi 5: \`timer_handler_count: 0\` despite \`sched_diag_idle_loops[0] = 43690\` (idle ran 43k times with \`daifclr + wfi\`).
- GIC state when sampled: \`HPPIR=30\` (timer IRQ pending, visible to non-secure), \`GICC.CTLR=1\`, \`PMR=0xf0\`, IRQ 30 priority \`0x80\` (passes priority mask), enabled, target CPU correct.
- **Direct UART writes at the very top of every EL1 vector** (\`el1_irq\`, \`el1_sync\`, \`el1_serror\` — before \`save_regs\`) never produced output during a session where tasks had \`DAIF.I=0\`. No exception of any type is being taken.
- The shell hangs at a deterministic point mid-\`uart_puts\` whenever tasks run with \`DAIF.I=0\`, consistent across retries — suggesting a specific hardware/firmware interaction, not an IRQ-storm / race.
- The \`pi5-baremetal-status.md\` claim of \"preemptive scheduling active\" is contradicted by the current behavior on \`origin/main\`.

## Suspected Areas

1. **GIC group configuration:** IGROUPR reads as 0 from non-secure (documented by earlier \`5a9b143\` investigation). Without the \`armstub8-2712.bin\` configuring GIC Group 1 at EL3, IRQs may be routed as Group 0 (secure) and thus rejected by non-secure EL1 — despite HPPIR being observable.
[...]

PR #636, opened 2026-05-06
sched: slm_preempt_point() — voluntary preemption points for COOP_PREEMPT busy-loop coverage (#635)

johnjezl:
## Summary

Adds `slm_preempt_point()` — a one-call cooperation point for in-tree CPU-bound loops on `COOP_PREEMPT` platforms (Pi 5, Jetson). Closes the documented "pure CPU-bound loop monopolizes its CPU" gap from the original Pi 5 cooperative-preemption resolution (PR #126, issue #99).

Track A of the Pi 5 true-preemption plan (#635); complementary to the longer-term hardware-IRQ-restoration work tracked under #134.

## What it does

- New header `kernel/include/preempt_point.h` defines the `slm_preempt_point()` macro:
  - **`COOP_PREEMPT` (Pi 5, Jetson):** expands to `slm_preempt_check_and_yield()` — reads `CNTPCT_EL0`, compares against the per-CPU 10 ms quantum, falls into `schedule()` only when the quantum has elapsed.
  - **non-`COOP_PREEMPT` (QEMU, x86-64):** expands to `((void)0)`. Call sites are unconditional.
- Implementation lives next to `coop_preempt_maybe_tick()` in `kernel/sched/sched.c`. Cost on the not-due path: ~5–8 instructions, branch not taken.
- Diagnostic counters `slm_preempt_point_calls` / `slm_preempt_point_schedule_calls` track fast/slow path entries.

## Call sites converted

- `kernel/tests/test_integration.c:delay()` — replaces the prior fixed 1024-iteration manual yield. Now reschedules once per quantum, not once per iteration budget.
- `kernel/ai_accel/hailo/hailo_control.c:wait_for_response` — 100 µs poll with up to 5 s timeout (50 000 iterations possible).
- `kernel/ai_accel/hailo/hailo_vdma.c` — both VDMA channel poll loops (descriptor-completion + wait-proc).

## Tests

- `test_slm_preempt_point_callable` (all platforms): 1 024 macro invocations must not fault.
- `test_slm_preempt_point_drives_schedule` (`COOP_PREEMPT` only): a 50 ms tight loop of macro calls must produce ≥ 4 schedule() entries (one per 10 ms quantum, allowing one missed boundary).

Both live in `kernel/tests/test_coop_preempt.c`.

## Docs

- `docs/scheduler.md` gains a new "Preemption Model" section: hardware-timer-IRQ vs `COOP_PREEMPT` vs `slm_preempt_point()`, plus the mandatory in-tree policy (loops that may iterate >1 000 times without yielding must call the macro).
- `kernel/CLAUDE.md` updated to point at the new policy from the existing COOP_PREEMPT section.

## Test plan

- [x] `make test` (QEMU ARM64) — all suites green
- [x] `make kernel PLATFORM=RASPI5` — clean build
- [ ] Deploy to pi-5-2 via labctl, `boot_test --count 10` — verify 10/10 with new test passing
- [ ] Verify `cpu` shell output shows `slm_preempt_point_calls` advancing under workload

### DIFF
--- a/docs/scheduler.md
+++ b/docs/scheduler.md
@@ -106,21 +106,27 @@ is cheap enough that "when in doubt, add it" is the right call —
 adding one preempt point per back-edge has no measurable cost on the
 hot path.
 
+**Hard restriction — DO NOT call `slm_preempt_point()` from inside a
+spinlock-held region.** SLM-OS's `spin_lock` and `spin_lock_irqsave`
+do not bump `preempt_disabled[]`, so the macro would fall into
+`schedule()` with the lock still held. A second task on the same CPU
+that contends the lock would spin forever (spin loops do not call
+`slm_preempt_point()`), single-CPU-deadlocking the kernel. Lock-held
+poll loops that need the scheduler to make progress are a separate
+problem — convert them to a sleep-based wait instead.
+
 Examples already following the policy:
 
-- `kernel/tests/test_integration.c:delay()` — loops `count` nops; one
-  preempt point per iteration.
-- `kernel/ai_accel/hailo/hailo_control.c:wait_for_response` — 100 µs
-  poll with up to 5 s timeout (50 000 iterations); one preempt point
-  per iteration. Lets the scheduler swap in another task while
-  firmware is still computing.
-- `kernel/ai_accel/hailo/hailo_vdma.c` — VDMA channel poll loops,
-  same structure as `wait_for_response`.
+- `kernel/tests/test_integration.c:delay()` — lock-free `nop` loop;
+  one preempt point per iteration.
 
-Reviewers looking at long loops should ask: *what's the worst-case
-iteration count, and is there at least one yielding call (or
-preempt-point) inside the loop body?* If neither, request the macro
-be added.
+Reviewers looking at long loops should ask three questions:
+1. *What's the worst-case iteration count?*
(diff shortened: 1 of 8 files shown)


johnjezl:
Hardware validation on pi-5-1 (pi-5-2 has a hardware issue producing zero serial output, so I switched to pi-5-1):

```
[TEST] kernel/tests/test_coop_preempt.c
  [PASS] test_cntpct_monotonic
  [PASS] test_pit_ticks_advances_over_time
  [PASS] test_timer_handler_count_advances
  [PASS] test_slm_preempt_point_callable
  [PASS] test_slm_preempt_point_drives_schedule
```

`test_slm_preempt_point_drives_schedule` is the COOP_PREEMPT-only soak test that exercises the macro on real Pi 5 hardware: a 50 ms tight loop of `slm_preempt_point()` calls drives ≥ 4 schedule() entries via the slow path. Confirms the quantum check fires on Pi 5.

The existing test suite is also clean (no regressions from the preempt-point insertion sites in `test_integration.c:delay()`, `hailo_control.c:wait_for_response`, and `hailo_vdma.c`).

`pi-5-2` produces zero serial output across 4 boot cycles regardless of kernel content — appears to be a serial / power / cable issue specific to that board. Logged here for visibility; happy to switch back when it's repaired.

PR #644, opened 2026-05-06
gpu/ga10b: evict inherited L2 lines on kexec inherit (#596)

johnjezl:
## Summary

Adds the documented NVIDIA L2-evict sequence (`FB_FLUSH → L2_FLUSH_DIRTY → L2_SYSMEM_INVALIDATE → FB_FLUSH`) to `ga10b_bringup_inherit()`, run once per kexec inheritance before any SLM-OS dispatch. Goal: kill the iter-1 stale-data race in #596 where the GPU returns Linux nvgpu's helper output instead of SLM-OS's fresh input.

## Why this matters

Sequence and poll-mask shape come from two independent references:
- `linux-nvgpu` rel-36 — `gv11b_mm_l2_flush` + `gk20a_mm_l2_flush` + `gk20a_mm_fb_flush` in `nvgpu-hal-mm-cache-flush_{gv11b,gk20a}_fusa.c`
- OGKM (`open-gpu-kernel-modules`) — `kmemsysCacheOp_GM200` for the order, comments for the GK20A sysmembar-before-L2-flush rule

Two non-obvious bits the prior attempts in this area (per the 2026-04-27 investigation summary) likely missed:
1. **Poll mask is BOTH `PENDING_BUSY=0` AND `OUTSTANDING_TRUE=0`.** Polling only `PENDING` exits before the op completes.
2. **Sysmembar (`FB_FLUSH`) ordering** — kmemsysCacheOp's comment cites the GK20A LTC IAS §5.5; same ordering applies to GA10B (Tegra Ampere).

`NV_UFLUSH_L2_SYSMEM_INVALIDATE` (= `_INVALIDATE_CLEAN_LINES` per the Hopper file's comment) is the specific op that purges *clean* cached reads — exactly the residue we inherit from nvgpu's MNIST helper.

## Test plan

- [x] Build clean (`make kernel PLATFORM=JETSON_ORIN_NANO`)
- [x] L2 evict runs cleanly on kexec inherit on jetson-nano-2 (6/6 fresh boot cycles)
- [x] No digit-5 stale-output signature observed: 0/11 dispatches across 6 cycles
- [ ] Statistical A/B vs. main, ~30 cycles each, via Lua FFI to match #596's original methodology — tracked separately
- [ ] Pair with Option A (double-dispatch warmup) — Bug A all-zero cold-start is a separate issue, dominated 5/6 first dispatches via shell path

## Caveats

This is a single-arm trial — no side-by-side fix-vs-no-fix comparison yet. The 17% pre-fix baseline came from a different methodology (likely Lua-via-FFI through the warmup), so "0/11 with fix" is consistent with the fix working but isn't proof. Marking draft until A/B stat run completes.

### DIFF
--- a/kernel/gpu/nvidia/ga10b_bringup.c
+++ b/kernel/gpu/nvidia/ga10b_bringup.c
@@ -897,10 +897,13 @@ int ga10b_bringup_pmu(struct ga10b_bringup *b)
  * gk20a_mm_l2_flush in nvgpu-hal-mm-cache-flush_gk20a_fusa.c — same
  * write-then-poll-both-bits pattern, same retry budget shape.
  *
- * Timeout: 100 retries × ~5 µs busy-wait = ~500 µs ceiling. nvgpu
- * uses 100 retries for FB_FLUSH, 200 for L2_INV, 2000 for L2_FLUSH;
- * 100 is sufficient for our use because we run this only at inherit
- * (no concurrent GPU traffic) so the op should complete in <10 µs.
+ * Timeout: 100 retries × ~5 µs busy-wait = ~500 µs per op ceiling.
+ * nvgpu uses 100 retries for FB_FLUSH, 200 for L2_INV, 2000 for
+ * L2_FLUSH; 100 is sufficient for our use because we run this only
+ * at inherit (no concurrent GPU traffic) so the op should complete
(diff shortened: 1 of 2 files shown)


johnjezl:
A/B test plan filed in #646 (will gate this PR's draft → ready).

PR #645, opened 2026-05-06
gpu/ga10b: double-dispatch first user call after kexec inherit (#596)

johnjezl:
## Summary

Adds a per-pipeline-kind "post-inherit throwaway" pass: on the first user-facing dispatch (`slm_gpu_run` / `slm_gpu_run_with_input`) after a fresh `ga10b_bringup_inherit()`, `launch_kernel` runs twice and the user sees the second dispatch's output. Complements PR #644 (L2 evict on inherit, now merged) — both target #596's iter-1 stale-data race.

## How it relates to the existing warmup

The warmup in `ensure_bringup` absorbs Bug A (very-first-dispatch returning all-zero logits). But the diagnostic captured in #596 showed the warmup output is *itself* bit-identical-stale across runs — it reads the same L2 lines that nvgpu's MNIST helper left cached pre-kexec. That's why the user-visible iter-1 stale-data rate was ~17% even with the warmup in place.

PR #644 evicts those L2 lines on inherit. This PR adds belt-and-suspenders for any lines that re-fill between inherit and the first user dispatch, by giving the user-call path one throwaway it discards.

## Cost

On the first user dispatch after kexec (per pipeline_kind), three GPU dispatches now run under the same `g_gpu_dispatch_lock` IRQ-off acquisition: existing warmup + the new throwaway + the user's real dispatch. Aligned with the file's own per-inference baseline:

- **Typical:** ~150–600 ms (3 × one MNIST inference at ~50–200 ms each)
- **Pathological** (hung GPU, every op hits `ga10b_submit_and_poll`'s 2 s timeout): bounded by 2 s × 8 ops × 3 dispatches ≈ **48 s** of IRQ-off

No effect on subsequent calls in the same kexec session. The lock-hold-time comment in `ensure_bringup` documents the new cost so a future maintainer can find it without re-deriving from the call chain.

## Test plan

- [x] Build clean (`make kernel PLATFORM=JETSON_ORIN_NANO`)
- [x] `make test` (QEMU ARM64) — passes
- [ ] Hardware smoke — deferred to the A/B run, which exercises this path via Lua FFI to match #596's original methodology
- [ ] A/B vs main + PR #644 alone (~30 cycles each via Lua) — tracked in #646

## Caveats

Marking draft until A/B confirms. The combined effect of (#644 + this) should drive the iter-1 stale-data rate to near zero; A/B will quantify the marginal contribution of each.

### DIFF
--- a/kernel/src/slm_ffi.c
+++ b/kernel/src/slm_ffi.c
@@ -625,13 +625,17 @@ int slm_gpu_dispatch_breaker_test_count(void)
  * overwrites `g_handoff`).
  *
  * Lock-hold-time note: callers run inside `g_gpu_dispatch_lock`
- * with IRQs disabled. This warmup doubles the worst-case lock-hold
- * on first dispatch — typical ~50-200 ms (one inference) but
- * `ga10b_submit_and_poll`'s 2 s per-op timeout means a hung GPU
- * could hold the lock + IRQs for ~16 s extra in pathological
- * cases. The user's dispatch already exercises this path (just
- * twice on first call); a future async-warmup-from-kernel_main
- * move would relieve the lock-hold cost. */
+ * with IRQs disabled. This warmup combines with the post-inherit
+ * throwaway dispatch (Option A for #596 — see
+ * `g_post_inherit_double_dispatch_pending`) to triple the
+ * worst-case lock-hold on first dispatch: warmup + throwaway +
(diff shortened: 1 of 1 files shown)


johnjezl:
A/B test plan filed in #646 (will gate this PR's draft → ready).

Issue #651, opened 2026-05-06
Jetson sched-MLP GPU dispatch wedges box under task burst (5 ms IRQ-off per assign_cpu)


johnjezl:
## Summary

`gpu use sched on` with the `ai_mlp` scheduler policy turns every `ai_mlp_assign_cpu` call into a synchronous GPU dispatch (`slm_gpu_run_sched_inference` in `kernel/src/slm_ffi.c`) holding `g_gpu_dispatch_lock` with `spin_lock_irqsave`. Per-call cost is ≈ 5 ms IRQ-off (CPU NEON is ≈ 50 µs). Six CPUs can all enter `ai_mlp_assign_cpu` from concurrent work-stealing / `task_create` paths; each waits on the global dispatch lock and adds another 5 ms of IRQ-off time when it gets in. Timer ticks, RX-stall watchdogs, and shell scheduling get starved and the system wedges.

## Reproducers (2026-04-27 jetson-nano-2)

Both wedged the kernel (telnet stops responding, serial console stops, networking goes unreachable, hard power-cycle required):

- `lua-admin /mnt/files/demo.lua` — script that creates several components/tasks; the script content is incidental, the task burst is what triggers it.
- A 2-iteration Lua loop calling `slm.task_create()` twice with 100 ms sleeps between them.

A single `slm.task_create('t', function() end)` runs cleanly (one 8-op dispatch on `kind=1` handoff, scanner finds it, returns to shell). The bug is concurrency-driven, not a single-dispatch issue.

## Root cause

Path: `ai_mlp_assign_cpu` (kernel/sched/ai/sched_ai.c:275) → `ai_assign_cpu_common` → `ai_schedule_mlp_via_device` → `forward_via_device` (kernel/sched/ai/ai_inference.c:280) → `slm_gpu_run_sched_inference` (kernel/src/slm_ffi.c:888) → acquires `g_gpu_dispatch_lock` IRQ-off → submits 8-op pushbuffer → polls semaphore → releases.

Each dispatch holds the lock with IRQs disabled for ~5 ms (typical) up to multiple seconds (pathological per `ga10b_submit_and_poll`'s 2 s/op timeout × 8 ops × concurrency). The scheduler invokes `assign_cpu` from any CPU during `scheduler_add_task` with no scheduler-wide lock above it, so all six CPUs serialise on `g_gpu_dispatch_lock`.

[...]

Issue #421, opened 2026-04-26
Jetson kernel image >16 MiB ceiling with AI_SCHED + ENABLE_NETWORKING


johnjezl:
Pre-existing build issue surfaced during admin-telemetry M1-M7 hardware verification on jetson-nano-2.

```
make kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON   EXTRA_KERNEL_CMAKE_ARGS="-DENABLE_NETWORKING=ON -DNET_TELNETD_AUTOSTART=ON -DNET_DHCP_AT_BOOT=ON -DGA10B_FIRMWARE_DIR=/home/john/jetson-ga10b-firmware"
...
ld: Kernel image too large!
```

The same failure reproduces on origin/main at e7a7d88 (pre-M1 baseline), so this is **not caused by the M1-M7 work**. The kernel-jetson.ld linker assertion is `__kernel_end < 0x81000000` (16 MiB).

**Workaround for hardware verification:** drop AI_SCHED. The build succeeds with just `-DENABLE_NETWORKING=ON -DNET_TELNETD_AUTOSTART=ON -DNET_DHCP_AT_BOOT=ON`. M1's `slm.sched_decision_rate()` returns nil under heuristic policy (which is the design); the rest of M2-M7 surface is exercised normally.

**Real fix options:**
1. Extend the Jetson linker map past 16 MiB. Need to check kexec handoff constraint and whether DTB / OP-TEE carveout overlaps.
2. Compress the AI_SCHED weight blobs (currently raw float arrays).
3. Make EMBED_DEMO_SCRIPTS off-by-default so demo .lua text doesn't pad the image when a build is tight.

Filing now so the constraint is captured; not a blocker for the current capstone path (CPU-MLP weights aren't required for the heuristic-policy demo).

PR #653, opened 2026-05-06
gpu/sched: rate-limit + try-lock the sched-MLP dispatch hot path (#651)

johnjezl:
## Summary

Closes the wedge documented in #651. `gpu use sched on` with `ai_mlp` was turning every `assign_cpu` into a 5 ms IRQ-off GPU dispatch, and a task burst across all six CPUs piled up enough waiters on `g_gpu_dispatch_lock` to starve timer ticks / RX-stall watchdogs / shell scheduling — full hard-wedge requiring a power-cycle.

## How

Two admission-control layers, both isolated to the sched path; MNIST and any other caller of `slm_gpu_run_with_input` are unchanged.

1. **Rate limiter** — system-wide atomic timestamp `g_sched_dispatch_last_ns`. Peek before any lock acquisition; if we're inside the 50 ms window from the previous *successful* dispatch, return -1 immediately. Updated only on rc==0 so failed dispatches don't waste the slot.
2. **Try-lock** — new `slm_gpu_try_run_with_input` is the same body as the blocking variant but with `spin_trylock` + manual `irq_save/restore`. If another CPU already holds the dispatch lock, return -1 instead of spinning IRQ-off behind a long dispatch.

`forward_via_device` (`kernel/sched/ai/ai_inference.c:280-289`) already has the silent CPU NEON fallback for negative rc, so no caller-side changes were needed.

## Effective change in IRQ-off load

| Scenario | Before | After |
|---|---|---|
| Steady state (1 dispatch every 50 ms) | ≤ 5 ms / 50 ms = 10 % of one CPU | unchanged |
| Burst (6 CPUs hammering `assign_cpu`) | up to 6 × 5 ms = 30 ms cumulative IRQ-off, all waiters serialise | ≤ 5 ms total per 50 ms window; other CPUs get -1 immediately and CPU NEON it |
| Pathological hang (one dispatch eats `ga10b_submit_and_poll`'s 2 s × 8 ops) | other CPUs spin IRQ-off behind it | other CPUs get -1 from try-lock and CPU NEON it |

## Test plan

- [x] Build clean for `JETSON_ORIN_NANO`
- [x] QEMU ARM64 `make test` passes
- [ ] Hardware reproducer on jetson-nano-2: `lua-admin /mnt/files/demo.lua` and `for i=1,2 do slm.task_create('t', function() end); slm.sleep_ms(100) end` under `gpu use sched on + ai_mlp` — both should complete cleanly instead of wedging the box (#651 acceptance).
- [ ] Sanity: a single-shot `slm.task_create('t', function() end)` still gets one GPU dispatch (rate limiter doesn't accidentally block the first one — `g_sched_dispatch_last_ns` starts at 0, `now - 0 >= RATE_LIMIT_NS` is true).

## Caveats

- 50 ms window is a starting point; tunable via `GPU_SCHED_DISPATCH_RATE_LIMIT_NS`. If a future workload wants tighter measurement granularity, the trade-off is more IRQ-off load.
- This does not address the underlying \"GPU dispatch is synchronous and blocking\" architecture. If a future product use-case needs the GPU result to actually drive the per-task decision, Option B from #651 (async queue with CPU fallback) is the path.

### DIFF
--- a/kernel/include/slm_ffi_test.h
+++ b/kernel/include/slm_ffi_test.h
@@ -20,6 +20,9 @@
 #ifndef SLM_FFI_TEST_H
 #define SLM_FFI_TEST_H
 
+#include <stdbool.h>
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -33,6 +36,16 @@ void slm_gpu_dispatch_breaker_test_record(int rc);
 /* Read the raw consecutive-failure counter. Atomically loaded so
  * tests can assert exact values mid-run. */
 int  slm_gpu_dispatch_breaker_test_count(void);
+
+/* Sched-MLP dispatch rate-limit predicate (#651).
+ *
+ * Returns true iff `now - last < RATE_LIMIT_NS`, i.e. the next
+ * dispatch should be rejected because the previous one finished
+ * inside the window. Available on every platform — the predicate
+ * itself is pure arithmetic — so the test runs cross-platform even
+ * though the call site (`slm_gpu_run_sched_inference`) is gated to
+ * Jetson. */
+bool slm_gpu_sched_dispatch_test_within_window(uint64_t now, uint64_t last);
 
 #ifdef __cplusplus
 }
--- a/kernel/src/slm_ffi.c
+++ b/kernel/src/slm_ffi.c
@@ -445,6 +445,30 @@ static struct ga10b_bringup g_bringups[GA10B_PIPELINE_KIND_COUNT];
 static bool g_post_inherit_double_dispatch_pending
     [GA10B_PIPELINE_KIND_COUNT] = { false };
 
+/* Sched-MLP dispatch rate limiter (#651).
+ *
+ * `gpu use sched on` with `ai_mlp` invokes `slm_gpu_run_sched_inference`
+ * from every `ai_mlp_assign_cpu` call — i.e. every task_create / steal
+ * across all six CPUs. Each dispatch holds `g_gpu_dispatch_lock` IRQ-
+ * off for ~5 ms; a task burst piles up dozens of waiters and starves
+ * timer ticks / RX-stall watchdogs / shell scheduling enough to wedge
+ * the box (telnet stops, serial stops, hard power-cycle required —
+ * 2026-04-27 jetson-nano-2 reproducer).
+ *
+ * The limiter is system-wide rather than per-CPU: only one GPU sched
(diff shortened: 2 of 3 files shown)


johnjezl:
## Hardware verification status — blocked by #421

Tried to verify on jetson-nano-2 but hit the kernel-image-too-large blocker:

\`\`\`
$ make kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON
[100%] Linking C executable slmos.elf
ld: Kernel image too large!
\`\`\`

The 32 MB ceiling in \`kernel/kernel-jetson.ld:135\` rejects the AI_SCHED build before linking. Largest contributors (from the .obj sizes):
- \`kernel/sched/ai/runtime_model.c.obj\` — 7.9 MB
- \`kernel/src/camera_mock_embed.S.obj\` — 2.5 MB
- \`kernel/arch/arm64/nvidia_ga10b_firmware.S.obj\` — 0.79 MB
- \`ai_weights_{mlp,ppo}.c.obj\` — 0.52 MB each

This is **#421**, not new. Worked around in the previous wedge-investigation session by some method I haven't replicated yet (possibly a custom build configuration that disabled some embedded blob).

Without \`AI_SCHED=ON\`, the \`ai_mlp\` scheduler policy isn't registered, so the new code path (\`slm_gpu_run_sched_inference\` → rate limiter → try-lock) is unreachable on hardware. Confirmed by attempting the test:

\`\`\`
slmos> sched policy ai_mlp
Unknown policy: 'ai_mlp'
\`\`\`

## What was verified

- ✅ Build clean for \`PLATFORM=JETSON_ORIN_NANO\` (default flags)
- ✅ \`make test\` (QEMU ARM64) passes
- ✅ Code review by inspection — see the diff
- ❌ Original wedge reproducer (Lua \`task_create\` × 2 with \`gpu use sched on + ai_mlp\`) — blocked

## Path forward

Two options to get hardware verification:
1. **Land #421 first** — fix the size ceiling (either bump the cap or trim the AI_SCHED runtime). Then rebuild + test.
2. **Stub out one of the large blobs** for an A/B test build — e.g. exclude camera_mock_embed and the AI runtime_model storage, just enough to fit AI_SCHED + the rate-limiter path under the 32 MB cap. Throwaway build for verification only.

Both are larger pieces of work than this PR. Recommendation: leave this PR draft until the size-ceiling blocker is addressed; the code itself is small, isolated, reference-backed (try-lock + atomic timestamp = standard kernel admission control patterns), and the QEMU build covers the build-correctness regression risk.

PR #656, opened 2026-05-07
diag: SCR_EL3 sentinel readout proves IRQ/FIQ clear takes effect (#134 Stage 2.5+)

johnjezl:
## Summary

End-to-end memory-sentinel debugging that proves our Pi 5 SCR_EL3
IRQ/FIQ clear in TF-A's \`setup_ns_context\` actually takes effect at
NS-EL1 entry — and is **not overridden** by any later EL3 path.

This rules out SCR_EL3 routing as the cause of the remaining #134 IRQ-
delivery gap, narrowing the search to the GIC group / \`ICC_IGRPEN0/1_EL1\`
side.

## Approach

Two sentinels in patched TF-A bracket the NS-context lifetime, write
to NC DRAM the kernel can read uncached:

- **Sentinel A** in \`setup_ns_context\` — magic + \`scr_el3\` after our
  IRQ/FIQ clear.
- **Sentinel B** in \`cm_prepare_el3_exit_ns\` — magic + \`scr_el3\` read
  back from the per-context buffer just before assembly \`el3_exit\`
  loads it via \`msr scr_el3, x18\`.

Kernel reads them via \`irqtest noirq\` (which already exists for #134;
extended here with a sentinel readout that runs before any DAIF
manipulation, so it's safe under all subcommands).

## Hardware result on pi-5-2

\`\`\`
--- SCR_EL3 sentinels (TF-A debug) ---
  Sentinel A (setup_ns_context):     magic=0xcafe0001 OK
                                     scr_el3=0x30731 (IRQ=0 FIQ=0)
  Sentinel B (cm_prepare_el3_exit_ns): magic=0xcafe0002 OK
                                     scr_el3=0x30731 (IRQ=0 FIQ=0)
  Verdict: A == B → no override between setup_ns_context and el3_exit.
\`\`\`

\`0x30731\` decodes as \`NS=1, RES1, HCE=1, RW=1, APK=1, API=1\`,
**\`IRQ=0\`, \`FIQ=0\`**. Both sentinels match.

## What this rules out / narrows

- ✅ Patched TF-A is loaded.
- ✅ The \`scr_el3 &= ~((1U << 1) | (1U << 2));\` in \`setup_ns_context\`
  takes effect.
- ✅ Nothing in TF-A between \`setup_ns_context\` and the assembly
  \`el3_exit\` re-writes the per-context SCR_EL3.
- ❌ \`irqtest noirq\` still reports zero IRQs/FIQs delivered.

So the remaining blocker is downstream of SCR_EL3 — most likely GIC
group register configuration (PPI 30 in Group 0 vs Group 1) or
\`ICC_IGRPEN0/1_EL1\` not enabled at NS-EL1 — not the security-state
routing.

## Build / deploy notes

- TF-A patch series regenerated as
  \`tools/tfa-patches/0001-…-IRQ-routing-patches.patch\` +
  \`tools/tfa-patches/0002-…-SCR_EL3-sentinels-with-EL3-NC-mmap-…patch\`.
- 0002 also adds the Pi 5 NC region (\`0xFFE00000\`+2 MB) to BL31's
  mmap as \`MT_NON_CACHEABLE | MT_RW | MT_SECURE\` — without that the
  sentinel store faults in EL3 with a synchronous external abort and
  \`plat_panic_handler\` infinite-loops before BL31 can print its banner.
  \`MAX_MMAP_REGIONS\` bumped 8 → 10 to match.
- Kernel side: slot defines in \`kernel/include/diag_pi5.h\`, readout
  in \`kernel/src/shell_sys.c\`'s \`cmd_irqtest\`. The readout runs first
  regardless of subcommand and decodes IRQ/FIQ bits from each value.

## Test plan

- [x] \`make test\` (QEMU) passes
- [x] \`make tfa-pi5\` builds 32870-byte bl31.bin with sentinel writes
      (verified via objdump — \`mov x0, #0xfec0; movk x0, #0xffe0\` +
      \`str w1, [x0]; str x19, [x0, #8]\` inlined into
      \`cm_init_my_context\`; identical sequence at \`0xfed0\` for B in
      \`cm_prepare_el3_exit_ns\`).
- [x] Pi 5 hardware (pi-5-2) boots cleanly to shell with the patched
      armstub.
- [x] \`irqtest noirq\` shows the sentinel readout above. Verdict
      A == B with IRQ=0 FIQ=0 reproduces.

## Status

Draft. Posting now to capture the Stage 2.5+ finding while we move on
to the GIC group / \`ICC_IGRPEN*\` investigation. Mark ready once the
follow-up GIC probe lands or the user is happy to merge as-is.


johnjezl:
## Sentinel C update — IGROUPR mystery resolved

Added a third sentinel that captures TF-A's own IGROUPR0 readback from Secure-EL3, plus a kernel-side readout in \`cmd_diag\`. Hardware result on pi-5-2:

\`\`\`
Sentinel C (TF-A IGROUPR readback): magic=0xcafe0003 OK
  TF-A view  IGROUPR[0]: pre=0xfe00ffff post=0xfe00ffff
  TF-A view  GICD_CTLR : 0x00000001
\`\`\`

Decoding \`0xfe00ffff\`:
- bits 0-15 (SGIs): Group 1 NS
- bits 16-24 (PPIs 0-8): Group 0 (Secure)
- bits 25-31 (PPIs 9-15, **including PPI 30 = timer**): Group 1 NS

So the pre-existing \`GICD_IGROUPR[0] writes SILENTLY DISCARDED\` diagnostic in the kernel's \`diag\` was a **read-side artifact**, not a write failure. NS-EL2 cannot read \`GICD_IGROUPR0\` (SGI/PPI range) on GIC-400 with security extensions — the Secure-only register returns 0 to NS reads. **PPI 30 (timer) is already in Group 1 NS.**

The IRQ-delivery gate for #134 is downstream of IGROUPR. Likely candidates next:
- \`GICD_CTLR.EnableGrp1NS\` not set at distributor level
- \`ICC_SRE_EL3.SRE\` locked → CPU forced to sysreg interface, Pi 5 GIC-400 only has MMIO

Continuing investigation.

johnjezl:
## Probe round 3 — GICD_CTLR ruled out

Added GICD_CTLR (NS view) to the runtime GIC dump in \`cmd_diag\`.

**Result on pi-5-2:**
\`\`\`
GICD_CTLR = 0x1  (NS view; bit0=EnableGrp1NS=1)  ← distributor forwards NS Group 1 ✓
\`\`\`

Combined with Sentinel C's TF-A view (\`GICD_CTLR : 0x00000001\` at TF-A time, Secure-view bit 0 = EnableGrp0), the actual register has **both** EnableGrp0 (TF-A) and EnableGrp1NS (kernel's \`gic_init\` write) set. Distributor is fully open for NS Group 1.

So the distributor is **not the gate**. PPI 30 is enabled, in Group 1 NS, pending in \`GICD_ISPENDR0\` — yet never reaches the CPU.

**Remaining lead:** \`ICC_SRE_EL3.SRE\`. If locked to 1, the Cortex-A76 listens on the sysreg ICC_* interface, but Pi 5's GIC-400 is MMIO-only — IRQs land in GICC and vanish. Original \`boot.S\` probe couldn't read \`ICC_SRE_EL2\` (traps to EL3 and silently halts the CPU). The sentinel infrastructure now lets us capture \`ICC_SRE_EL3\` directly from EL3.

johnjezl:
## Probe round 4 — ICC_SRE branch ruled out (and a misdirected first attempt)

First Sentinel D attempt read \`ICC_SRE_EL3\`/\`EL2\`/\`EL1\` from EL3. Hung the board. Trimmed to EL3-only — still hung. Rolled TF-A back to the Sentinel C state to recover.

Root cause of the hang is visible in the existing \`ID_AA64PFR0_EL1 = 0x1100000010111112\` capture: bits 27:24 (the GIC field) read as **0**, meaning the Cortex-A76 in BCM2712 **does not implement the GIC system-register interface**. With \`GIC == 0\`, \`ICC_SRE_EL{1,2,3}\` accesses are UNDEFINED at every EL — which is why even an EL3 read of \`ICC_SRE_EL3\` faulted.

\`diag\` now decodes the GIC field directly from the boot.S id_aa64pfr0 capture:
\`\`\`
ID_AA64PFR0_EL1.GIC = 0 (NOT implemented — CPU has no sysreg ICC_*)
  >>> ICC_SRE_EL* are UNDEFINED on this CPU. SRE branch
      of #134 is moot — IRQ delivery has to flow through
      legacy GICC_* MMIO unconditionally.
\`\`\`

So the SRE branch is dead. The IRQ-delivery gate is somewhere else on the MMIO path. Suspects remaining:
- **GIC-400 NS priority shift**: Secure-written PPI 30 priority of 0x80 might not be visible/comparable correctly through the NS PMR=0xf0.
- **DAIF.I in tasks**: kernel runs all tasks with DAIF.I=1 in the cooperative model. \`irqtest\` (without \`noirq\`) is the canonical test, but it's risky — hangs on success because schedule() runs from IRQ context without SECONDARY_PREEMPT.

Sentinel D slots are left vacant for future use; \`diag_pi5.h\` now documents the reservation.

johnjezl:
## Probe round 5 — IRQ delivery confirmed; SECONDARY_PREEMPT path needs work

### Linux diff findings (offline reading)

Compared against \`drivers/irqchip/irq-gic.c\` and \`drivers/clocksource/arm_arch_timer.c\` from the reference cache. Two concrete differences:
1. **Linux clears \`GICC_ACTIVEPRIO[0..3]\` in \`gic_cpu_if_up\`** — we never touch it.
2. **Linux writes IPRIORITYR for IRQs 0-31 from NS state with \`0xa0\`** in \`gic_cpu_config\` — we rely on TF-A's earlier Secure-side \`0x80\` write.

Added captures for both to the runtime \`diag\` dump (commit \`be6066a3\`).

### Hardware result

\`\`\`
GICC_ACTIVEPRIO[0..3]: 0x0 0x0 0x0 0x0  (clean)
IPRIORITYR[PPI 30] (NS view) = 0x80   PMR = 0xf0
  delivers under PMR if pri < PMR  →  yes (priority allows delivery)
\`\`\`

Both Linux-derived hypotheses are clean — neither was the gate. Every visible GIC bit is correctly configured for delivery.

### IRQ delivery confirmed

Ran \`irqtest\` (without \`noirq\`). Trace stopped at checkpoint **3** (right at \`msr daifclr, #2\`):
\`\`\`
... sentinel readout ...
WARNING: this probe will hang the system if a timer IRQ fires
... 5-second pause ...
 1 2 3
\` ← (no further output for 20+ seconds — board hung)
\`\`\`

This is the **canonical success signature**: the moment IRQs were unmasked, the pending timer IRQ fired and the handler called \`schedule()\` from IRQ context, corrupting the abandoned exception frame. Hardware IRQ delivery has been working for some time — we just never tested it because the schedule-corruption hazard was kept gated behind \`SECONDARY_PREEMPT=ON\` (default OFF).

### Bottom line for #134

The infrastructure to make hardware preemption work is:
- Patched TF-A (this PR) → SCR_EL3 + IGROUPR + GIC routing all correct ✓
- Existing trampoline (kernel/sched/preempt.c, vectors.S \`resched_trampoline\`) → defers \`schedule()\` to task context ✓
- \`SECONDARY_PREEMPT=ON\` build flag → activates the trampoline ✓

### Open: SECONDARY_PREEMPT=ON hardware bring-up

Built with \`SECONDARY_PREEMPT=ON\`, deployed, booted cleanly, ran \`irqtest\` — sentinel readout printed, then the shell wedged before the "irqtest probe:" header. The trampoline path was previously documented as "functional but inert" because IRQs didn't deliver; actually exercising it on hardware surfaced bugs that need to be tracked separately.

This PR's contribution is **the diagnostic infrastructure that pinpoints the actual gate**. The trampoline-bring-up work is its own follow-on issue (file separately under #134).

johnjezl:
Filed follow-on issue #672 for the SECONDARY_PREEMPT trampoline bring-up. PR #656 stays scoped to the diagnostic infrastructure that proved hardware IRQ delivery works.

Issue #683, opened 2026-05-07
Pi 5: move SLM-OS to EL2 with VHE so NS IRQ delivery works (#672 follow-on)


johnjezl:
## Problem

NS-EL1 IRQ delivery on Pi 5 (BCM2712 / GIC-400) is broken regardless of which timer PPI we pick. Confirmed in #672 by switching the timer between PPI 30 (CNTP, NS Phys) and PPI 27 (CNTV, virtual) — both produce identical symptoms:

- The GIC enables and pends the IRQ correctly. From NS-EL1, `GICC_HPPIR` returns the actual IRQ id (not 1022), proving the IRQ is in Group 1 NS and visible to NS.
- The CPU's IRQ pin is asserted (proven indirectly: disabling the timer via `CNTP_CTL=0` or disabling the PPI via `GICD_ICENABLER0` lets a subsequent `daifclr` complete; otherwise it hard-locks).
- But the IRQ exception vector at NS-EL1 is **never entered** — `vec->irq` stays 0 across all CPUs forever, even when DAIF.I is cleared.
- `daifclr` clearing the I bit while the GIC pin is asserted hard-locks CPU 0 with no exception delivered.

Linux on Pi 5 doesn't hit this because it boots into **EL2 with VHE** and uses **PPI 26 (Hyp Physical Timer)**, taking IRQs at EL2 via VBAR_EL2. That code path is what Pi firmware validates; NS-EL1 IRQ routing is not. See `arch_timer_select_ppi()` in `drivers/clocksource/arm_arch_timer.c` — the first branch (`is_kernel_in_hyp_mode()` → `ARCH_TIMER_HYP_PPI`) is what Pi 5 hits, every time.

So the root cause is almost certainly some combination of GIC-400 SoC integration / Pi firmware GICC routing that's only exercised at EL2; the kernel-side observable symptoms are an empirical artifact of \"Pi firmware doesn't deliver NS-EL1 IRQs.\"

## Goal

Run SLM-OS at NS-EL2 with VHE so that:
1. The kernel takes IRQs at EL2 via VBAR_EL2 — the path Linux uses and Pi firmware actually validates.
2. The timer becomes PPI 26 (Hyp Physical Timer); CNTP_*_EL1 register accesses are silently redirected to CNTHP_*_EL2 by VHE.
[...]

Issue #672, opened 2026-05-07
Pi 5 SECONDARY_PREEMPT trampoline wedges shell when actually exercised (#134 follow-on)


johnjezl:
## Background

PR #656 (\`stage25/scr-sentinel-debug\`) confirmed that hardware timer IRQs DO deliver to NS-EL1 on Pi 5 with our patched TF-A — proven by the canonical \`irqtest\` hang signature at checkpoint 3 (right at \`msr daifclr, #2\`). The remaining gate for #134 is the \`schedule()\`-from-IRQ-context safety, which is the \`SECONDARY_PREEMPT=ON\` trampoline path.

The trampoline (\`kernel/sched/preempt.c::maybe_arm_resched_trampoline\` + \`kernel/arch/arm64/vectors.S::resched_trampoline\`) was previously documented as "functional but inert because IRQs didn't deliver." Now that they do, exercising the path on hardware surfaces a real bug.

## Reproduction

1. \`make kernel-clean && make kernel PLATFORM=RASPI5 PI5_IRQ_DIAG=ON SECONDARY_PREEMPT=ON\`
2. Deploy to pi-5-2 (with patched TF-A from PR #656).
3. Boot — comes up cleanly, shell starts.
4. Run \`irqtest\` (without \`noirq\`).

## Observed

- The Sentinel A/B sentinel readout prints normally.
- Shell wedges immediately after the verdict line — the next print (\`irqtest probe:\` header) never appears.
- Subsequent shell input (newline, \`help\`) gets no response.
- Power-cycle recovers.

With \`SECONDARY_PREEMPT=OFF\`, the same kernel boots and shell is responsive — only the trampoline path is broken.

## Expected

\`irqtest\` should run to completion: print all 7 checkpoints, increment \`vec->irq\` (via \`DIAG_BUMP_VEC\`), and report \`RESULT: IRQ DELIVERED\`. The shell prompt should return.

## Likely culprits

- \`maybe_arm_resched_trampoline\` may be re-arming the trampoline mid-context-switch, clobbering \`orig_elr\` / \`orig_spsr\`. The current guard is \`preempt_disabled[cpu]\`, but every shell print path holds \`uart_lock\` (IRQ-disabled), so the timing window is narrow.
- The \`PSR_I\` force in \`tf->spsr\` should keep IRQs masked through the trampoline prologue, but may interact poorly with the way \`schedule()\` re-enables IRQs on the new task.
[...]

PR #685, opened 2026-05-07
#683 PR 2: Pi 5 boot at EL2 with VHE

johnjezl:
## Status

**WIP — silent boot wedge on pi-5-1 hardware.** Builds clean across QEMU/RASPI5/JETSON, QEMU `make test` passes, but Pi 5 produces zero UART output post-flash. Intended hand-off to a fresh agent for hardware debug.

## Goal

PR 2 of the #683 EL2/VHE refactor. Move Pi 5 from EL1h to EL2h with VHE (HCR_EL2.E2H=1, TGE=1, RW=1) so hardware IRQs route through `VBAR_EL2` — the path Linux on Pi 5 uses and Pi firmware actually validates. NS-EL1 IRQ delivery on BCM2712 / GIC-400 is broken regardless of which timer PPI is selected (#672 closing summary), so running at EL2 is the only path forward.

The plan is `docs/pi5-el2-vhe-plan.md` (added in this PR's first commit). PR 1 audit results are in the same file under "PR 1 audit results".

## Commits in this PR

1. `f3fd26e6` — `docs(#683): EL2/VHE refactor plan + PR 1 audit + close predecessor docs`
2. `6ef4ac40` — `feat(#683/pi5,wip): boot at EL2 with VHE — silent boot regression on hw`

## What the code does

* `kernel/arch/arm64/el_regs.h` (new) — `KERN_ELR` / `KERN_SPSR` macros, gated on `PLATFORM_RASPI5`. ELR/SPSR are *not* in the VHE auto-redirect set, so they need explicit `*_EL2` operands at EL2h.
* `kernel/arch/arm64/vectors.S` — 6 sites switched from `elr_el1`/`spsr_el1` to the macros (save_regs prologue, restore_regs epilogue, resched_trampoline orig_elr/orig_spsr restore).
* `kernel/arch/arm64/boot.S` — Pi 5 EL2 block stays at EL2 with VHE instead of `eret`-ing to EL1. RMW pattern on HCR_EL2 (matches the Jetson EL2/VHE block at lines 669-672). `CPACR_EL1`, `SCTLR_EL1`, `CNTHCTL_EL2` writes happen post-E2H so they hit the right (redirected) registers. Removes the legacy `cptr_el2 = 0x33ff` and `hstr_el2 = xzr` writes (E2H=0-layout trap setup) and the SPSR/ELR/eret sequence that dropped to EL1.
* `kernel/arch/arm64/smp_boot.S` — secondaries also stay at EL2 on Pi 5 (matches primary). Heterogeneous EL was unsafe to test, so SMP got pulled into PR 2 from the original PR 3 scope. Jetson's existing EL2/VHE path is preserved.

QEMU and other non-RASPI5/non-JETSON ARM64 platforms are unaffected — the EL2 code is gated on `CurrentEL == 8` at runtime *and* on `PLATFORM_RASPI5` at compile time.

## Hardware regression

After flashing this build to **pi-5-1** and power-cycling: **zero UART output** across multiple boots over 25-30s windows. Reverted to base `main` on the same SD card and the shell came up cleanly, so the SBC + flash path are healthy. The wedge is in this commit's code.

### Ruled out
1. **Build/binary integrity.** `objdump -d build/kernel/slmos.elf` confirms the HCR_EL2 RMW emits `orr x10, x10, x11 ; msr hcr_el2, x10 ; isb`, with `x11` loaded from a literal pool whose value is `0x488000000` = bit 34 (E2H) | bit 31 (RW) | bit 27 (TGE).
2. **vectors.S macro expansion** — irrelevant to early boot, no exception is taken before the wedge.
3. **SBC health** — same SD card boots base `main` cleanly to shell.

### Suspect order

**A. Pre-existing HCR_EL2 bits getting OR'd in.** Pi 5 firmware may leave HCR_EL2 with `VM=1` (stage-2 enable), `IMO`/`FMO=1`, or other bits. The original code did an unconditional overwrite (`hcr_el2 = 1<<31`) which cleared all of these. The new RMW preserves them. If `VM=1` is set without stage-2 page tables, the first memory access after the E2H+TGE flip could fault silently (no vector installed). The PI5_IRQ_DIAG slot at +0x00 captures HCR_EL2 pre-write — but the kernel that reads it is wedged.

**B. `CPTR_EL2` / `HSTR_EL2` firmware default trap bits.** Removing the legacy writes leaves those at firmware values. Probably not it (no FP code in the gap), but worth confirming.

**C. `CNTHCTL_EL2` layout under E2H=1.** ARM ARM says with E2H=1 the register switches to EL1 CNTKCTL layout. Same bit positions for what we set (0+1), so writing `3` should be correct under both. But if EL2 view has RES1 bits the EL1 view doesn't, writing a bare `3` could be UNPREDICTABLE. Worth comparing against Linux's `__init_el2_timers` in `arch/arm64/kernel/head.S`.

**D. Heterogeneous secondary state.** Pi 5 firmware spawns secondaries via TF-A; PSCI CPU_ON wakes them. That's not called until `smp_init()` runs much later, so the secondary path shouldn't be exercised in this window. Sanity check: nothing in `smp_boot.S` runs prematurely.

The most likely fix is **A** — try unconditional `msr hcr_el2, <E2H|TGE|RW>` (overwrite, like the original `RW=1`-only code) instead of RMW. If that boots, we know firmware leaves a problematic bit set and we need to mask it explicitly rather than preserve it.

## Suggested debug next steps

1. **UART byte trace.** Use the existing `STAGE25_TRACE_CHAR` macro pattern from `vectors.S` (writes a single byte to RP1 PL011 DR at 0x1f00030000 via movz/movk; doesn't depend on a literal pool). Drop one character right after the `isb` at the end of the EL2/VHE block (`boot.S:~422`), another at `primary_cpu` entry (`boot.S:733`), and a third just before `bl kernel_main` (`boot.S:853`). Whichever character is the last to appear bisects the wedge to a ~50-line region.
2. **Try unconditional HCR_EL2 overwrite.** Replace the `mrs/orr/msr` block with `mov x10, #0 ; movk x10, ... ; msr hcr_el2, x10` setting only E2H|TGE|RW. If that boots, hypothesis A is confirmed and the fix is to mask out problematic firmware bits.
3. **Capture firmware HCR_EL2.** Add a `STAGE25_TRACE_CHAR` immediately after the `mrs x16, hcr_el2` capture so we can see, byte by byte, what firmware left there.
4. **Compare against Linux's Pi 5 EL2 setup.** `arch/arm64/kernel/head.S` `init_kernel_el` and `__init_el2_*` macros are the reference.

## Repro

```bash
git checkout 683/pr2-pi5-el2-vhe
make kernel-clean && make kernel              # QEMU, builds clean
make test                                     # passes
make kernel PLATFORM=RASPI5                   # builds clean
make kernel PLATFORM=JETSON_ORIN_NANO         # builds clean
# Hardware:
labctl claim_sbc pi-5-1 30 "#683 PR 2 debug"
labctl sdwire_update pi-5-1 --partition 1 \
       --copies build/kernel/slmos.bin:kernel_2712.img --reboot true
labctl serial_capture pi-5-1 --timeout 25
# Result: 0 lines captured → silent boot wedge
```

## Acceptance for merge

- [ ] Pi 5 hardware boot reaches the shell on pi-5-1.
- [ ] `diag` shell command shows `CurrentEL = 0xC` (EL2h).
- [ ] HCR_EL2 post readback in the diag snapshot has bits 27, 31, 34 set.
- [ ] `boot_test --count 10` 10/10 on pi-5-1, parity with base `main`.
- [ ] QEMU `make test` continues to pass.
- [ ] QEMU and Jetson builds remain clean.

After merge, follow-on PRs from `docs/pi5-el2-vhe-plan.md`:
- PR 3: timer to PPI 26 (Hyp Physical Timer) — once #683 PR 2 lands, this is what unblocks hardware IRQ delivery.
- PR 4: userspace EL0 → EL2 SVC routing.

## References

- Parent issue: #683 (move SLM-OS to EL2/VHE on Pi 5)
- Predecessor: #672 (closed — NS-EL1 IRQ delivery broken regardless of PPI)
- Original blocker: #134 (Pi 5 hardware timer IRQ delivery)
- Plan + PR 1 audit: `docs/pi5-el2-vhe-plan.md`
- Disconfirming evidence: `docs/pi5-stage25-irqtest-findings.md`

## PR 1 findings (summary; full audit in `docs/pi5-el2-vhe-plan.md`)

Inventory of every `mrs`/`msr` access in scope (boot.S, vectors.S, smp_boot.S, context.S, mmu.S, exceptions.c, timer.c, platform.h, preempt.c, vmm.c) classified for the EL2/VHE move:

| Disposition | Sites | What it means |
|---|---|---|
| **VHE-OK** | 23 | `*_EL1` operand silently redirects to `*_EL2` under E2H=1; no source change. |
| **rewrite-EL2** | 6 | Operand must change to `*_EL2` (all `elr_el1`/`spsr_el1` — handled via `KERN_ELR`/`KERN_SPSR`). |
| **drop-or-rewire** | 9 | Block restructure (Pi 5 EL2-drop path in boot.S/smp_boot.S, timer.c CNTV). |
| **platform-guard** | 4 | Per-platform compile gate (HCR_EL2 setup, TIMER_IRQ, ICC_*_EL1). |
| **no-change** | 12 | Identifier-style register (MIDR/MPIDR/ID_*/CCSIDR/CSSELR/CNTFRQ/CNTPCT). |

Net: only 6 sites genuinely need explicit EL2 register-name rewrites; the 23 VHE-OK sites compile and execute correctly via auto-redirection. The bulk of the refactor is the two structural changes (boot.S + smp_boot.S EL1-drop blocks).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -930,6 +930,8 @@ set(KERNEL_INCLUDES
 # resorting to relative includes.
 if(PLATFORM STREQUAL "X86_64")
     list(APPEND KERNEL_INCLUDES ${CMAKE_SOURCE_DIR}/kernel/arch/x86_64)
+else()
+    list(APPEND KERNEL_INCLUDES ${CMAKE_SOURCE_DIR}/kernel/arch/arm64)
 endif()
 
 # Build kernel ELF — include lwIP when networking is enabled
--- a/docs/pi5-armstub-track-c.md
+++ b/docs/pi5-armstub-track-c.md
@@ -1,6 +1,10 @@
 # Pi 5 Track C — Custom EL3 Armstub for True Preemptive Multitasking
 
-**Status:** Stage 1 (minimal armstub source + build) and Stage 2 (custom TF-A with `SCR_EL3` + GIC patches) both landed. Stage 2.5 (kernel-side `DAIF.I` unmask) — open.
+**Status:** **Closed.** Stage 1 (armstub source + build) and Stage 2 (custom TF-A with `SCR_EL3` + GIC patches) both landed and validated on hardware. Stage 2.5 disconfirmed the original "one-line `DAIF.I` unmask" framing — see [`pi5-stage25-irqtest-findings.md`](pi5-stage25-irqtest-findings.md) for the differential probing that ruled out NS-EL1 IRQ delivery on Pi 5.
+
+The actual production fix is tracked in [#683](https://github.com/SLM-OS/SLM-Operating-System/issues/683) — move SLM-OS to EL2 with VHE so IRQs route through `VBAR_EL2` (the path Pi firmware actually validates). See [`pi5-el2-vhe-plan.md`](pi5-el2-vhe-plan.md) for the refactor plan.
+
+This document is retained for the Stage 1 + Stage 2 work product (TF-A patches in `tools/tfa-patches/`, `make tfa-pi5`, custom `bl31.bin` deploy path) which remain useful even under the EL2/VHE direction.
 
 **Issue:** [#134](https://github.com/SLM-OS/SLM-Operating-System/issues/134) — restore hardware timer IRQ delivery on Pi 5.
 
@@ -215,49 +219,43 @@ rarely runs because the shell + `net_pump` keep CPU 0 busy.
 the CPU never holds an unmasked IRQ state long enough to take
 the pending timer IRQ.
 
-## Stage 2.5 — kernel-side: per-task `orig_elr`/`orig_spsr` (open)
+## Stage 2.5 — original framing was wrong (resolved into #683)
 
-This stage was originally framed as "unmask `DAIF.I` in tasks" — a
-one-line change. **Hardware testing on pi-5-2 with the irqtest probe
-in this PR proved that's only half the story.**
+This stage was originally framed as "unmask `DAIF.I` in tasks plus
+move `orig_elr`/`orig_spsr` into `struct task`" — small kernel-side
+work. The differential probing on `issue/672/secondary-preempt-bringup`
+(see [`pi5-stage25-irqtest-findings.md`](pi5-stage25-irqtest-findings.md)
+and #672's closing summary) **disconfirmed** that framing.
 
-**Verified via `irqtest`:** with the Stage 2 TF-A loaded and
-`DAIF.I` briefly cleared from a shell command, the system **hangs
-immediately** — meaning the timer IRQ DOES deliver to the IRQ
-vector. Stage 2's TF-A patches *are* working. The hang itself is
-proof.
+What the probing established:
 
-**What's actually broken:** Pi 5's existing `SECONDARY_PREEMPT`
-trampoline (`kernel/sched/preempt.c`, PR #98) saves `orig_elr` /
-`orig_spsr` in **per-CPU NC slots**, not per-task. When
-preempt-during-preempted-task happens (which it does as soon as
-hardware IRQs are firing every 10 ms across multiple ready tasks),
-the inner preemption clobbers the outer's save slots, and the
-outer task's eret target is corrupted.
+- Stage 2's TF-A patches **do** run and **do** put PPI 30 in Group 1 NS.
+  `GICC_HPPIR` from NS-EL1 returns 30 (not 1022) — proof that the GIC
+  routing is correct.
+- The GIC's `nIRQ` pin to CPU 0 is asserted. Disabling the timer
+  (`CNTP_CTL=0`) or the PPI (`ICENABLER0`) unblocks `daifclr`; the
+  pin assertion is real.
+- **But the EL1 IRQ vector is never entered** — `vec->irq` stays 0
+  across all CPUs. `daifclr #2` at NS-EL1 with the pin asserted
+  hard-locks CPU 0 with no exception delivered.
+- PPI 27 (CNTV virtual) tested as the alternative: identical wedge.
+  The blocker is **not PPI-specific**; NS-EL1 IRQ delivery is broken
+  in general on this Pi 5 / BCM2712 / GIC-400 firmware.
 
-The fix is moving `orig_elr` / `orig_spsr` into `struct task`. This
-is a small structural change (~20 lines in preempt.c, vectors.S,
-task.h) plus regression validation. It belongs in Stage 2.5 because
-it's the load-bearing change for activating real preemption.
+The trampoline (`orig_elr`/`orig_spsr` in `struct task`) was never
+exercised because the IRQ exception is never delivered to enter it.
+The trampoline infrastructure remains structurally sound but inert.
 
-After Stage 2.5:
+**Path forward:** [#683](https://github.com/SLM-OS/SLM-Operating-System/issues/683)
+moves SLM-OS to EL2 with VHE so IRQs go through `VBAR_EL2` — the
+path Linux on Pi 5 uses and Pi firmware validates. See
+[`pi5-el2-vhe-plan.md`](pi5-el2-vhe-plan.md) for the refactor plan.
 
-1. **Unmask `DAIF.I`** in `task_entry_trampoline` (gated on
-   `SECONDARY_PREEMPT=ON`). One line.
-2. **Activate `SECONDARY_PREEMPT=ON`** for Pi 5 default builds.
-3. **Activate the Stage 2 custom TF-A** (place `bl31.bin` →
-   `armstub8-2712.bin` on the boot partition, add `armstub=` to
-   `config.txt`).
-4. **Boot test:** `boot_test --count 10` on pi-5-2. Each boot
-   should advance per-CPU IRQ counters > 0.
-5. **Multi-core test acceptance:** the five originally-failing
-   tests should pass via real hardware preemption rather than via
-   the Tier 2/3 auto-recovery.
-
-The `irqtest` shell command (`PLATFORM_RASPI5 + PI5_IRQ_DIAG`
-only, in `kernel/src/shell_sys.c`) stays in this PR as the
-diagnostic that pinned the per-task `orig_elr` requirement and
-will continue to pin Stage 2.5's success.
+The `irqtest` shell command (`PLATFORM_RASPI5 + PI5_IRQ_DIAG`,
+`kernel/src/shell_sys.c`) stays as the diagnostic harness that
+pinned the EL1-IRQ-delivery wedge and will pin #683's success
+(after #683's PR 4 lands, `irqtest` should report `RESULT: IRQ
+DELIVERED`).
 
 ## Acceptance criteria for closing #134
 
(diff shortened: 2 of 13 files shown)


johnjezl:
## Pi-5-2 reproduction: 10/10 boots, EL2/VHE active

Picked this up on `pi-5-2` (the agent-assigned preempt board, also the board referenced in `docs/pi5-el2-vhe-plan.md` PR-2 acceptance). The wedge described in the PR body does **not** reproduce here.

### Hardware test results

```
boot_test --runs 10 (commit 6ef4ac40, RASPI5 build)
Result: 10/10 boots successful (100%)
Average boot time: 13.8s
```

Diag dump from `diag el2` post-boot:

```
EL2 register snapshot (boot.S, issue #99):
  CurrentEL at entry:     0x8 (EL2)
  HCR_EL2 pre:            0x80000000          ← only RW=1, no firmware bits
  HCR_EL2 post (written): 0x488000000         ← E2H(34) | RW(31) | TGE(27)
  CNTHCTL_EL2:            0x3
```

Firmware on pi-5-2 leaves HCR_EL2 with **only** `RW=1` set (`0x80000000`). The RMW pattern this PR adopted is preserving zero firmware-set bits, so it produces the same value an unconditional overwrite would: `0x488000000`. Boot reaches the shell, scheduler starts, timer arms, all 4 CPUs come online, MNIST preload kicks off — full kernel functionality at EL2h with VHE.

### Banner accuracy fix (commit 413774fc just pushed)

The `[INFO] Running at EL1 on Raspberry Pi 5` line was hardcoded for non-Jetson/non-x86 platforms. Replaced with a runtime read of `CurrentEL` + `HCR_EL2.E2H` so the banner now correctly reports `[INFO] Running at EL2 (VHE) on Raspberry Pi 5`. QEMU (still EL1) and x86-64 paths are unchanged. Pushed as a follow-on commit on this branch.

### Disposition of PR-body acceptance criteria

| Criterion | Status on pi-5-2 |
|---|---|
| Pi 5 hardware boot reaches the shell | ✅ 10/10 |
| `diag` shows CurrentEL = 0xC (EL2h) and HCR_EL2.E2H = 1 | ✅ (`CurrentEL at entry = 0x8`, post-flip banner says EL2 with VHE) |
| HCR_EL2 post readback has bits 27, 31, 34 set | ✅ (`0x488000000`) |
| `boot_test --count 10` 10/10 on hardware, parity with main | ✅ on pi-5-2 |
| QEMU `make test` continues to pass | ✅ |
| QEMU and Jetson builds remain clean | ✅ (RASPI5 + QEMU verified locally) |

### Why the asymmetry between pi-5-1 and pi-5-2

The PR body's hypothesis A (firmware HCR_EL2 difference) is the most plausible explanation. pi-5-2's firmware leaves `HCR_EL2 = 0x80000000` (just RW). If pi-5-1's firmware leaves additional bits set (e.g. `VM=1` for stage-2, `IMO/FMO=1` for IRQ trapping), the RMW preserves them and the resulting HCR_EL2 value can wedge the CPU when E2H+TGE are added on top. Lab boards likely differ in EEPROM version (cf. `pi5_eeprom_findings.md`).

### Recommended next steps

Options for moving forward:

1. **Merge PR-2 on pi-5-2 verification, file pi-5-1 wedge as a follow-up.** The plan doc's acceptance criterion is explicitly pi-5-2 ("Acceptance: pi-5-2 boots to shell. `diag` shows `CurrentEL = 0xC` (EL2h) and `HCR_EL2.E2H = 1`."). Both met. The pi-5-1 wedge becomes a post-merge investigation that could land independently.
2. **Add an HCR_EL2 byte-trace diagnostic before merge.** Plant a `STAGE25_TRACE_CHAR` sequence around the HCR_EL2 read that nibble-dumps the value to UART. Whoever has pi-5-1 access can flash and immediately see what firmware leaves there. Low-cost, gated under `PI5_IRQ_DIAG`.
3. **Switch to unconditional overwrite (PR-body hypothesis A fix).** Replace the RMW with `mov + msr` setting only `E2H | RW | TGE`. This would mask any problematic firmware bits. Risk: if firmware does leave a bit set that the kernel actually needs (unlikely — there's no `VM`-stage-2 setup elsewhere), unconditional overwrite breaks it.

I'd lean toward **(1) merge + follow-up** since pi-5-2 verification is solid and the wedge doesn't block the next PR's work (timer to PPI 26). Happy to add (2) if you want pi-5-1 diag in this PR.

Test transcript captured; raw output available on request.

Tasks: completed `Reproduce PR #685 silent boot wedge on pi-5-2` (wedge did not reproduce); created `Investigate pi-5-1 vs pi-5-2 firmware differences` to track the residual.

johnjezl:
## Review-fix delta — commit b6019861

Addressed actionable findings from `/slmos-review`:

### Code changes

- **boot.S, smp_boot.S — restored `msr hstr_el2, xzr`.** Pi 5 firmware enters with unknown HSTR_EL2 contents. Inheriting them risks trapping VHE-redirected `*_EL1` accesses (sctlr_el1, cpacr_el1) to EL2 with no handler installed. PR description's suspect B for the pi-5-1 wedge.
- **boot.S, smp_boot.S — switched HCR_EL2 from RMW to unconditional overwrite.** RMW preserves firmware-set bits (potentially `VM=1`, `IMO/FMO=1`, etc.) which may wedge the CPU after the E2H flip on at least one Pi 5 firmware variant. Now matches the Jetson EL2 block. PR description's suspect A.
- **el_regs.h — included `PLATFORM_JETSON_ORIN_NANO`.** Jetson runs at EL2/VHE per smp_boot.S; KERN_ELR/KERN_SPSR now resolve to `*_EL2` there too. Added C-friendly string-form macros `KERN_ELR_NAME` / `KERN_SPSR_NAME` for inline-asm use.
- **panic.c — migrated to KERN_ELR_NAME/KERN_SPSR_NAME.** Panic register dumps now capture the live exception state on Pi 5/Jetson instead of dormant EL1-bank registers.
- **user_entry.S — migrated to KERN_ELR/KERN_SPSR.** No production caller on Pi 5 today (only test_syscall.c), but pre-emptively correct for the day EL0 user tasks are wired up.

### One reviewer suggestion was wrong, kept

Reviewer flagged the pre-flip `msr cnthctl_el2, x10` as redundant ("value 3 correct under both layouts"). I removed it and pi-5-2 silently wedged with zero UART output. Restored after empirical bisect — the pre-flip write is **load-bearing** on Pi 5. Comment in boot.S now records this; not removing without hardware bisect would be the right rule going forward.

### Hardware validation

`boot_test --runs 10` on pi-5-2:
```
9/10 boots successful
1 fail = lab-infra Kasa auth flake (not code)
All real boots succeed; banner: "Running at EL2 (VHE) on Raspberry Pi 5"
```

Pi 5 / Jetson / QEMU all build clean. `make test` passes.

### Remaining

The pi-5-1 wedge is expected to be resolved by the HSTR_EL2 + unconditional-HCR_EL2 changes. Needs hardware verification on pi-5-1 (the original failing board) — I only have access to pi-5-2. If pi-5-1 still wedges after this commit, the next step is the byte-trace diagnostic the PR body proposed (plant `STAGE25_TRACE_CHAR` markers around the HCR_EL2 read to dump firmware state to UART pre-flip).

PR #690, opened 2026-05-07
hailo: #682 defer ENABLED to runmodel (disconfirmed)

johnjezl:
## Summary

Flag-gated experiment to test whether moving `CHANGE_STATUS(ENABLED)` + CFG pull from load time to runmodel time (so they fire milliseconds before the IN `avail` bump instead of seconds) unblocks fw's IN ch=2 `proc` advance. **Disconfirmed on pi-5-1.**

Adds `HAILO_DEFER_ENABLED_TO_RUN` (default **OFF**). When ON, `context_switch_load` stops after the 4 SET_CONTEXT_INFO RPCs and sets `slot->enabled_deferred=true`. The first `hailo_backend_run` on that slot then issues `CHANGE_STATUS(ENABLED)` + CFG pull immediately before its boundary submit.

## Hypothesis source

#682 issue body §3: *"If fw expects [APPLICATION_CHANGE_INTERRUPT] to fire only after host signals input ready, the order of events between host arming the channel and host poking ENABLED may need inversion."*

## Refactor

- Extracts the ENABLED + CFG pull block from `context_switch_load` (~150 lines) into a new file-static helper `hailo_apply_enabled_and_cfg_pull()`. Reads `ccw_num_avail_{0,1}` + `ccw_has_second_channel` from the slot rather than load-local variables. Behavior preserved when flag is OFF — identical RPC sequence, stage codes, WIRE_DEBUG instrumentation.
- Adds `bool enabled_deferred` field to `struct hailo_model_slot`.
- Adds the gated deferred-call site at the top of `hailo_backend_run`, one-shot per slot (clears the flag immediately on success).

## Verdict on pi-5-1 (2026-05-07)

**DISCONFIRMED.** With ENABLED + CFG pull running milliseconds before the IN avail bump (instead of seconds at load time), fw's IN ch=2 stalls **identically**:

| signal | load-time ENABLED | runmodel-time ENABLED |
|---|---|---|
| `base_post` (host) | `0x00022801` | `0x00022801` |
| `proc` after 500 ms | 0 | 0 |
| `dev base` mirror | `0x00002c01` | `0x00002c01` |
| desc[0..7] status | `0x00` | `0x00` |
| `submit_and_wait` rc | `-4` | `-4` |
| CPU_ECC events | seen | seen (same payload) |

The IN channel's gating mechanism is not about when ENABLED is issued.

## Notable side observation

When the deferred path runs at runmodel time, CFG ch=0 immediately reports `proc=0x006e006d` (109/110) **before** the runmodel-time ENABLED writes execute. Either fw was processing the dynamic block in the background even without ENABLED (which contradicts our existing comment "before ENABLED the channel is all zeros, writes don't stick"), or ch=0 is reporting stale state from a prior inference. Worth a clean-board re-test if this becomes load-bearing for a future experiment.

## Test plan

- [x] Build with default `HAILO_DEFER_ENABLED_TO_RUN=OFF` — refactored code path runs identically (helper called inline at end of load)
- [x] Build with `HAILO_DEFER_ENABLED_TO_RUN=ON` — load stops after SET_CONTEXT_INFO(DYNAMIC), reports "context-switch load OK", no ENABLED issued
- [x] Boots on pi-5-1, hailo probe/boot/load/runmodel all parse
- [x] Runmodel emits the deferred ENABLED + CFG pull bisect markers in order before IN submit
- [x] Compare run trace before/after — no difference in proc advance, dev mirror, desc status, ECC notifications

## Notes

- **Not a #682 fix.** Closing the issue is gated on a working `runmodel`. PR is documented disconfirmation + a re-orderable hook.
- Independent of #688 (IRQ-cycle) and #689 (settle gap). Either of the prior PRs can land first; this one rebases cleanly on top of either.
- Default OFF — load-time activation matches Linux's order more closely and is the correct default. The flag is preserved so the alternative ordering is selectable for future A/B without re-editing source.

## Related #682 PRs (all disconfirmed)

- #688 — IRQ-cycle (`IMASK_HOST` disable→re-arm around fw boot, default ON for Linux parity)
- #689 — OUT→IN settle gap (default OFF)
- This — Defer ENABLED to runmodel (default OFF)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,28 @@ if(HAILO_D3HOT_AT_BOOT)
     add_compile_definitions(HAILO_D3HOT_AT_BOOT=1)
 endif()
 
+# Defer CHANGE_STATUS(ENABLED) + CFG pull from context_switch_load to the
+# first hailo_backend_run on the slot. #682 hypothesis (issue body §3):
+# fw expects host to signal input ready before the dynamic context's
+# APPLICATION_CHANGE_INTERRUPT action fires, so moving the ENABLED
+# trigger closer to the avail bump may shorten the window in which fw
+# sits idle waiting for nothing.
+#
+# Verdict on pi-5-1 (2026-05-07): DISCONFIRMED. With ENABLED+CFG pull
+# moved from load to runmodel (firing milliseconds before the IN
+# avail bump instead of seconds), fw's IN ch=2 still stalls
+# identically: proc=0 throughout, dev base=0x00002c01, all desc
+# status=0x00. The gating mechanism for IN ch=2 is not about
+# ENABLED timing.
+#
+# Default OFF — load-time activation matches Linux's order more
+# closely and is the correct default. Kept gated so the alternative
+# can be re-enabled for comparison without re-editing source.
+option(HAILO_DEFER_ENABLED_TO_RUN "Defer CHANGE_STATUS(ENABLED) from load to first run (#682, disconfirmed)" OFF)
+if(HAILO_DEFER_ENABLED_TO_RUN)
+    add_compile_definitions(HAILO_DEFER_ENABLED_TO_RUN=1)
+endif()
+
 # Pi 5 IRQ-delivery diagnostics (issue #99 Phase 1). Adds EL2 register
 # snapshot in boot.S, per-CPU per-vector exception counters in vectors.S,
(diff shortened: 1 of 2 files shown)


johnjezl:
Closing per session review (2026-05-07): disconfirmed-experiment knob, default OFF. Issue #682 captures the experiment trail. Closing to avoid disconfirmed-knob foot-guns in the tree.

PR #753, opened 2026-05-09
arch/arm64: save FP/SIMD state in IRQ trap frame

### LINKED ISSUE #750: Jetson: user-mode task faults when hardware timer preempts (JETSON_HW_TICK=ON only)
## Summary

After PR #746 + #749 landed hardware-IRQ delivery to NS-EL2 on Jetson Orin Nano (verified working: SLM-OS receives PPI 26 / CNTHP), enabling the actual hardware-preempt scheduler config wedges a user-mode task on the first preemption:

- \`JETSON_HW_TICK=OFF\` (coop default): boots all the way to \`slmos>\` shell, lwIP up, mnist preload kicked, 100% clean.
- \`JETSON_HW_TICK=ON SECONDARY_PREEMPT=ON COOP_PREEMPT=OFF\`: kernel takes a user-mode-task fault, prints region map + call stack, then BL31 RAS handler hits an SError on a stale GIC redistributor frame and powers off the core.

A/B-bisected this session on jetson-nano-2.

## Symptoms (full sequence with HW_TICK=ON)

\`\`\`
[INFO] Boot successful
[INFO] Running at EL2 (VHE) on Jetson Orin Nano
...
[INFO] Initializing VMM...
[INFO] Enabling MMU...
[IRQ] Unhandled IRQ 198 — masking at GIC      <-- xudc, fixed by #749
��t23x_adc_core_average_get: ACC req clear ack timeout    <-- BPMP firmware noise on Cortex-R5 (cosmetic)
FPU exception in user mode                                  <-- BPMP firmware noise (32-bit ARMv7 register dump)
r0_usr  0x0000186a   ... r12_usr  0xc0000000  spsr 0x60000030  <-- BPMP-side
...
00: base: 50000000 size: 00800000 ...                      <-- SLM-OS region map dump (panic path)
global worker@0x50248824 sp 0x5018ff08 stack: 5018e000 - 5018fffc
call stack:
sp 0x5018ff08 pc 0x50022724
sp 0x5018ff48 pc 0x5002d550
sp 0x5018ffc8 pc 0x5004550c
sp 0x5018ffe0 pc 0x500694f6
sp 0x5018fff8 pc 0xaaaaaaaa                                <-- stack canary, normal unwind end
��ERROR: RAS Uncorrectable Error in IOB ... ADDR=0x800000000f5fff40   <-- in GICR space, frame 13 (Jetson has 6 CPUs)
ERROR: sdei_dispatch_event returned -1
ERROR: Powering off core
\`\`\`

## Confirmed not the cause

- ❌ The xudc IRQ — fixed by #749, the storm is silenced before this fault hits.
- ❌ BPMP-side fault — runs on a separate Cortex-R5, cosmetic TCU mux noise on the same serial console, no causal link to the SLM-OS fault.
- ❌ Stock-BL31 vs patched-BL31 — patched BL31 (#746) is fine on coop boot.
- ❌ Pre-existing — coop boot is clean, so this is introduced by HW-IRQ-driven scheduling.

## Likely culprits, in priority order

1. **\`switch_to\` not preserving user-mode (EL0) state when invoked from IRQ context.** SLM-OS's cooperative scheduler always schedules from kernel context (yield-point driven); the HW-IRQ path is the first time \`schedule()\` runs with EL0 caller state in SPSR_EL2/ELR_EL2. If the save side doesn't account for that, the user task's PC/SP/x-regs are partially clobbered on resume → fault at the \`pc\` address shown.
2. **FPU state save/restore.** \`resched_trampoline\` saves x0-x18 + x29/x30 but not Q0-Q31 / FPCR / FPSR. If the interrupted user task had dirty FPU state, the trampoline+schedule path either corrupts it or trips a re-entry trap. (Pi 5 worked because its EL2/VHE preemption was tested with \`net_pump\`-style integer workloads; user-mode FPU is a Jetson-side delta if the user blob has float math.)
3. **GIC RAS error at 0x0F5FFF40.** Address falls in the GICR region for a CPU index past the populated range. Likely a panic-time GIC walk using \`MAX_CPUS\` (8) instead of \`cpu_count\` (6 on Jetson). Probably a downstream consequence of the panic, not the original cause.

## Reproduction

\`\`\`
make kernel PLATFORM=JETSON_ORIN_NANO JETSON_HW_TICK=ON SECONDARY_PREEMPT=ON COOP_PREEMPT=OFF
scp build/kernel/slmos.elf root@192.168.4.5:/boot/slmos-test.elf
ssh root@192.168.4.5 sudo /usr/local/bin/slmos-kexec /boot/slmos-test.elf
labctl serial_capture jetson-nano-2 --timeout 120
\`\`\`

A coop-baseline rebuild (drop the three flags) reproduces clean boot.

## Out of scope

- BPMP-side noise (separate issue if it's worth fixing — likely just a slmos-kexec teardown improvement to quiesce BPMP's watchdog before handoff).

## Proposed first investigation step

Compare \`switch_to\` on the IRQ-from-EL0 path vs the coop yield-from-kernel path:
1. \`make kernel PLATFORM=JETSON_ORIN_NANO JETSON_HW_TICK=ON SECONDARY_PREEMPT=OFF COOP_PREEMPT=OFF\` — timer fires + IRQ-driven \`schedule()\` from ISR (no trampoline). If this also faults at the same PC, the problem is in \`schedule()\`/\`switch_to\` save side itself, not the trampoline.
2. If it doesn't fault (just hangs the kernel, since schedule-from-ISR is unsafe on real hardware), the trampoline path is the differentiator and \`resched_trampoline\` register save coverage is the suspect.

johnjezl:
## Summary

- Defensive structural fix. Extends `save_regs`/`restore_regs` in `kernel/arch/arm64/vectors.S` to save the full q0-q31 + FPCR + FPSR file on every IRQ entry. Frame grows 272 → 800 bytes.
- Closes a real silent-FPU-corruption class for any IRQ-driven preemption that swaps tasks. Without it, AAPCS64 lets C callees in the IRQ path (el1_irq_handler, maybe_arm_resched_trampoline) freely clobber q0-q7 + q16-q31 — and the post-eret trampoline → `schedule()` → `switch_to(old, new)` then captures the clobbered values as `old->context.v[]`.
- **NOT the fix for #750** (#752 is). This was investigated and tested on hardware as a candidate for #750; the failure mode persisted, ruling out FPU corruption as the trigger. The change is structurally correct and orthogonal to that bug; merging is a separate decision.

## Why this matters even though it didn't fix #750

In coop-preempt mode, `schedule()` runs from `yield()` at task context — live q-regs == task's q-regs at `switch_to`, so the AAPCS-clobber problem doesn't materialize. With hardware-driven preemption (Pi 5 #742 already shipping; Jetson #750 once the NULL-check fix lands), every quantum runs through the trampoline → schedule path with C-clobbered q-regs in the live registers when `switch_to` snapshots them. Coverage in `save_regs`/`restore_regs` is the simplest place to fix it — the same eager-save shape `switch_to` already uses for cooperative context switches.

## Cost

~20 extra instructions per IRQ entry/exit (16 stp/ldp pairs + 2 mrs/msr + 1 str/ldr) and 528 extra bytes of stack frame. Acceptable universal cost — there is no FPU-trap mechanism in this kernel that would let us defer to first-use.

## Test plan

- [x] QEMU: `make kernel` clean, `make test` passes (one unrelated Hailo flake).
- [x] Jetson: `JETSON_HW_TICK=ON SECONDARY_PREEMPT=ON COOP_PREEMPT=OFF` builds clean. Hardware boots to shell when paired with #752 (this PR alone doesn't fix #750).
- [ ] Pi 5 hardware regression: not yet retested; the `save_regs`/`restore_regs` path is shared, so ~20-instruction-per-IRQ regression risk in the Pi 5 hot path. Recommend a `bench` round on pi-5-2 before merge.

### DIFF
--- a/kernel/CLAUDE.md
+++ b/kernel/CLAUDE.md
@@ -212,6 +212,20 @@ The `MAX_TASKS * STACK_SIZE` worst-case PMM budget is now 16 MB
 (64 tasks × 256 KB). Comfortable on every shipping platform; bump
 the assertion in `config.h` if `MAX_TASKS` ever grows past 64.
 
+**ARM64 IRQ trap-frame budget (PR #753):** Every IRQ entry on
+ARM64 allocates `TRAP_FRAME_ALLOC` = **800 bytes** on the
+interrupted task's kernel stack — 264 bytes for x0-x30 + ELR +
+SPSR plus 528 bytes for the full FP/SIMD register file
+(q0-q31 + FPCR + FPSR). The FP slots exist so AAPCS64 caller-save
+clobbers inside `el1_irq_handler` /
+`maybe_arm_resched_trampoline` can't corrupt the interrupted
+task's q-regs (see save_regs in `kernel/arch/arm64/vectors.S`
+for the rationale). Combined with the trampoline's own
+~192-byte frame and any nested exception, expect a peak of
+~1.5 KB consumed for IRQ machinery alone. Negligible against
+the 256 KB per-task budget; flagged here so future stack-size
+audits factor it in.
+
 ---
 
 ## Stack-Overflow Guard — save-side SP-range assertion (May 2026)
--- a/kernel/arch/arm64/vectors.S
+++ b/kernel/arch/arm64/vectors.S
@@ -10,6 +10,7 @@
 
 #include "el_regs.h"
 #include "cpu_id_asm.h"
+#include "trap.h"
 
 .section .text
 
@@ -87,14 +88,16 @@ exception_vectors:
  * Macro to save all general-purpose registers AND FP/SIMD state.
  * Creates a trap frame on the stack.
  *
- * Frame layout (800 bytes, 16-byte aligned):
- *   [sp, #0..239]    x0-x29 (15 stp pairs)
- *   [sp, #240]       x30
- *   [sp, #248..263]  ELR, SPSR
- *   [sp, #264..271]  padding (keeps q-region 16-byte aligned)
- *   [sp, #272..783]  q0-q31 (16 stp pairs of q,q at stride 0x20)
- *   [sp, #784..791]  FPCR
- *   [sp, #792..799]  FPSR
+ * Frame layout (TRAP_FRAME_ALLOC bytes, 16-byte aligned — defined
+ * symbolically in kernel/include/trap.h, with _Static_asserts that
+ * pin every offset below to the C view):
+ *   [sp, #0..239]                  x0-x29 (15 stp pairs)
+ *   [sp, #240]                     x30
+ *   [sp, #248..263]                ELR, SPSR
+ *   [sp, #264..271]                padding (keeps q-region 16-byte aligned)
+ *   [sp, #TRAP_FRAME_OFF_QREGS+0..511]  q0-q31 (16 stp pairs at stride 0x20)
+ *   [sp, #TRAP_FRAME_OFF_FPCR]     FPCR
+ *   [sp, #TRAP_FRAME_OFF_FPSR]     FPSR
  *
  * AAPCS reserves x18 as the "platform register" — toolchains that
  * generate code for a runtime which uses x18 as a TLS / shadow-call
@@ -126,14 +129,20 @@ exception_vectors:
  *      the corrupted q-regs are restored. Old task continues with
  *      garbage FPU state.
  *
- * Same shape as switch_to in context.S (eager save of full FP/SIMD
- * file). Cost: 16 stp pairs + 2 mrs + 1 stp = ~20 instructions +
- * 528 bytes of stack per IRQ on every ARM64 path. Acceptable
- * universal cost; there is no FPU-trap mechanism in this kernel
- * that would let us defer the save to first-use.
+ * Lazy save (CPACR_EL1.FPEN trap-on-FP-use, then save on first-use)
+ * is NOT implemented here. SLM-OS already eagerly saves the full
+ * q0-q31 + FPCR + FPSR file in `switch_to` (kernel/arch/arm64/
+ * context.S), so making the IRQ path also eager keeps both paths in
+ * the same model. Switching to lazy save would be a coordinated
+ * change across both paths and is out of scope for #750.
+ *
+ * Cost: 16 stp pairs + 2 mrs + 2 str = ~20 instructions + 528 bytes
+ * of stack per IRQ on every ARM64 path. Pi 5 hardware bench (#753):
+ * `bench irq` min jitter +112 ns (3% over baseline), p50/p99 within
+ * noise.
  */
 .macro save_regs
-    sub     sp, sp, #800            /* GPRs + ELR/SPSR + pad + q-regs + FPCR/FPSR */
+    sub     sp, sp, #TRAP_FRAME_ALLOC   /* GPRs + ELR/SPSR + pad + q-regs + FPCR/FPSR */
     stp     x0, x1, [sp, #0]
     stp     x2, x3, [sp, #16]
     stp     x4, x5, [sp, #32]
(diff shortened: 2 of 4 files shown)


johnjezl:
## Pi 5 hardware bench (pi-5-2, 2026-05-09)

`PLATFORM=RASPI5 SECONDARY_PREEMPT=ON COOP_PREEMPT=OFF` (so every quantum exercises the IRQ → trampoline → schedule path the FPU save covers). Bench command suite: `bench context`, `bench irq`, `bench smp`. Multiple runs of `bench irq` to capture variance.

### Results

| Metric | Baseline (no FPU save) | This PR (FPU save) | Delta |
|---|---|---|---|
| `bench context` p50 | 2074 ns | 2092 ns | +18 ns (+0.9%) |
| `bench context` p99 | 2407 ns | 2426 ns | +19 ns (+0.8%) |
| `bench context` mean | 2148 ns | 2123 / 2161 ns | within noise |
| `bench irq` min (4×3 runs) | 3666 ns | 3778 ns | +112 ns (+3%) |
| `bench irq` typical avg | 4630-4682 ns | 4025-4689 ns | within noise |
| `bench irq` 20 µs outlier | **occurs ≥ 3/4 baseline runs** | **occurs ~1/3 FPU runs** | not FPU-induced |
| `bench smp` cross-CPU dispatch | 3/3 complete | 3/3 complete | no regression |

### Notes

- The +112 ns minimum IRQ jitter increase is the cost of 16 stp + 16 ldp + 4 mrs/msr ≈ ~70-150 ns at Pi 5's ~2.4 GHz, consistent with the predicted overhead.
- The 19-20 µs jitter spikes appear in **both** kernels — they are background variance (likely cache eviction or scheduler-housekeeping coinciding with the bench), **not** induced by this PR.
- Boot reaches `slmos>` cleanly. `timdiag` shows clean 100 Hz hardware preemption. `bench context` runs to completion without anomalies. `bench smp` cross-CPU dispatch unaffected.

### Conclusion

No measurable Pi 5 performance regression. Recommend merging.

johnjezl:
## /slmos-review feedback addressed (commit a828b731)

All actionable items from the [review](https://github.com/SLM-OS/SLM-Operating-System/pull/753) addressed:

### Warnings (must fix)
- **Drift guards**: `trap.h` now exposes `TRAP_FRAME_OFF_QREGS / FPCR / FPSR / ALLOC` as `#define`s; `vectors.S` `#include`s `trap.h` and references symbolic offsets at every site. `_Static_assert`s in `trap.h` pin layout invariants (struct size, q-region 16-byte alignment, FPCR/FPSR offsets relative to q-region, ALLOC coverage + 16-byte alignment). A future change drifts → compile error.
- **Regression test**: new `test_irq_save_regs_preserves_fp_simd` in `kernel/tests/test_syscall.c`. Loads a unique 128-bit pattern into each of q0-q31 via inline asm, busy-loops on `pit_ticks` until a timer IRQ fires (CNTPCT-bounded 200 ms deadline), then verifies the pattern element-by-element on the round-trip. Pit_ticks advances only from inside `timer_handler`, called from `el1_irq_handler` — so an observed advance proves `save_regs`/`restore_regs` ran with our pattern live in q-regs. QEMU `make test`: **2337 PASS / 1 unrelated FAIL** (pre-existing Hailo `test_control_identify_arms_imask_once` flake on `origin/main` baseline too).
- **Stack-budget doc**: `kernel/CLAUDE.md` "Task Stack Sizing" section gains an explicit note that ARM64 IRQ entry now consumes 800 B per IRQ.

### Suggestions
- Save/restore comments explicitly note that `x0` is reused as scratch for FPCR/FPSR `mrs/msr` — original caller value at offset 0 (or restored from offset 0 on exit).
- Save block records the lazy-vs-eager FPU rationale: no `CPACR_EL1.FPEN` trap-on-FP-use; `switch_to` in `context.S` also eagerly saves the full file, so the IRQ path is consistent.
- Symbolic offsets (`TRAP_FRAME_OFF_QREGS + N`) replace hard-coded `272/304/.../752` everywhere.

### Builds verified
- QEMU virt: clean, `make test` passes (modulo unrelated flake)
- Jetson Orin Nano `JETSON_HW_TICK=ON SECONDARY_PREEMPT=ON COOP_PREEMPT=OFF`: clean
- Raspberry Pi 5 `SECONDARY_PREEMPT=ON`: clean

Diff: +368 / -79 across 4 files.

johnjezl:
## /slmos-review pass 2 feedback addressed (commit 5174191b)

All three suggestions resolved:

1. **Deadline overflow-safety comment** — added inline. Documents that on every shipping ARM64 platform CNTFRQ is 19.2–54 MHz so `freq/5 ≤ ~11M`; CNTPCT post-boot is bounded by uptime in counter ticks (no realistic platform wraps for ~10^10 years). Notes the (small) regression class that would need rework if a future platform reports CNTFRQ anywhere near 2^61.

2. **Bit-mixed per-register test pattern** — index now placed in bits[31:24] of the low half and bits[63:56] of the high half, so no two q-regs can produce identical (low, high) pairs under any permutation of save/restore order:
   ```
   low  = 0xDEADBEEF_<i>5A5A5A
   high = 0x<i>5A5A5A_CAFEBABE
   ```
   Defeats the thin class of swap aliases the prior `0xA5/0x5A`-with-index-in-bottom-byte pattern theoretically left open. Test still PASSes (`[PASS] test_irq_save_regs_preserves_fp_simd`).

3. **Definitive FP-clobber hook in el1_irq_handler** — filed as **issue #754**. Reviewer themselves explicitly classified this as a follow-up ("invasive enough that bundling it with #753 isn't worth blocking on"). The hook would make the test fail deterministically when `save_regs`/`restore_regs` regress, even when the C handler happens to be integer-only — strict improvement over the current "necessary but not sufficient" coverage.

Diff: +20 / -7 in `kernel/tests/test_syscall.c`. QEMU `make test` clean modulo unrelated Hailo flake.

PR #781, opened 2026-05-11
scripts: capture-hailo-trace.sh orchestrator (PR C)

johnjezl:
## Summary
- One-command Pi 5 trace capture: builds the kernel, claims the lab board, flashes kernel + `cmdline.txt` with the requested `hailo_trace.phase=` / `hailo_trace.mech=` tokens, captures serial through boot, optionally fires post-boot shell commands, and drops the capture into `~/slmos-ref/derivatives/slmos-traces/slmos-<scenario>-<board>-<date>.txt`.
- Wraps the manual labctl workflow used during PR A's hardware verify (#287). Single deliverable: 226-line bash script under `scripts/`.

## Stacking
Based on `feat/hailo-trace-framework` (#780) — the script needs the trace masks + cmdline-parser PR A introduces. PR target is `feat/hailo-trace-framework` so the diff view shows only the script. When PR A merges, rebase onto main is a no-op.

## Why this exists
During PR A's hardware verify I ran the build → power-off → sdwire-update → power-cycle → serial-capture → flip-mask-via-shell → serial-send → cleanup cycle by hand. Doing it again next time would mean repeating that sequence; bottling it into a script means the archive at `~/slmos-ref/derivatives/slmos-traces/` grows naturally whenever someone needs a fresh capture, and the file metadata header (date, board, kernel commit, masks) is consistent across captures.

## Args
```
--phase <list>         comma list, or 'all'/'off'
--mech  <list>         same shape
--shell-cmd "<cmd;…>"  optional post-boot commands (semicolon-separated)
--scenario <name>      tag in the output filename
--board <name>         labctl SBC name (default pi-5-1)
--skip-build           reuse build/kernel/slmos.bin
--keep-cmdline         leave cmdline.txt on the card (default: removed)
--duration <minutes>   claim duration (default 30)
--out-dir <path>       override archive location
```

Example reproducing PR A's bootphase capture verbatim:
```
scripts/capture-hailo-trace.sh \
    --phase link,fw_boot,postboot \
    --mech  pci,rpc,irq \
    --shell-cmd "hailo probe; hailo boot" \
    --scenario bootphase
```

## Idempotency contract
On any exit path (success / failure / Ctrl-C / SIGTERM) the cleanup trap:
1. Removes `cmdline.txt` from the SD card (unless `--keep-cmdline`)
2. Releases the labctl claim
3. Preserves partial captures if the run failed midway

Re-running the script is always safe — it never assumes prior card state.

## Validated
- [x] `bash -n` syntax check
[...]

PR #780, opened 2026-05-11
hailo/trace: runtime-gated boundary trace framework (PR A)

johnjezl:
## Summary
- Runtime-gated boundary trace framework for the Hailo path. Two bitmasks — `hailo_trace_phase_mask` and `hailo_trace_mech_mask` — gate emission of one line per MMIO/PCI-cfg/RPC/IRQ/DMA/busy-wait transaction. Hot-path guard inlines to a single load+test+branch when disabled; both masks default to 0u so kernels built without explicit defaults are byte-identical to the silent-baseline path.
- Three control surfaces: build-time `HAILO_TRACE_PHASE_DEFAULT` / `HAILO_TRACE_MECH_DEFAULT` (covers pre-shell phases like Phase 1 PCIe link train), `cmdline.txt` → `/chosen/bootargs` tokens `hailo_trace.phase=...` / `hailo_trace.mech=...` (no rebuild needed), and the `hailo trace [...]` shell subcommand for post-shell iteration.
- Output format matches the existing Linux `hailort-traces/` patch where mechanisms overlap (MMIO + PCI cfg) so cross-side captures diff cleanly.

## What this enables
- Boot-time capture of Phase 1 (PCIe link train) MMIO/PCI-cfg without instrumenting individual call sites — every transaction routes through `pi5_read32`/`pi5_write32` and the 6 `pcie_config_*` accessors which now emit when armed.
- Live diff against the Linux `~/slmos-ref/derivatives/hailort-traces/` archive for cross-validation.
- Phase-tagged transaction logs that let `[trc] phase=postboot mech=PCI WR16 bdf=01:00.0 off=0xfc val=0x0000200b` get cross-referenced to lifecycle doc §1 Phase 3 step 10 (the PMCSR D0→D3hot transition).

## Commits
1. **`71fec1ca` framework foundation** — `hailo_trace.h` + `hailo_trace.c`, no call sites wired. ~640 LOC.
2. **`bd5bf1a1` wire MMIO/PCI/RPC/IRQ/DMA hooks + phase boundaries** — instruments all transaction points and inserts `hailo_trace_set_phase()` at 6 lifecycle entries (LINKUP in `main.c`, FW_BOOT in `hailo_boot`, POSTBOOT post-disarm, MODEL_LOAD in `inference_device_hailo`'s `load_model`, INFERENCE in `backend_run`, TEARDOWN in `signal_driver_shutdown`). ~500 LOC.
[...]

PR #783, opened 2026-05-11
scripts/capture-hailo-trace.sh: Pi 5 boundary-trace capture orchestrator (PR C)

johnjezl:
## Summary

Standalone capture orchestrator for the `hailo_trace` boundary-trace framework merged in #780. Single command builds an SLM-OS kernel for Pi 5 with the requested trace masks armed via `cmdline.txt`, deploys to a labctl-managed SBC, captures serial output through boot + (optionally) a follow-on shell command, and saves the result to `~/slmos-ref/derivatives/slmos-traces/` with a date-stamped filename.

Re-creating #781 because the original was auto-closed when `feat/hailo-trace-framework` was squash-merged and deleted. Same head branch (`feat/hailo-trace-capture-script`), now retargeted to `main`. Same commit (`43430ca8`), rebased on top of the merged framework.

## Files

- `scripts/capture-hailo-trace.sh` (226 lines, `bash -n` + `shellcheck 0.9.0` clean)

## Args

| Arg | Effect |
|---|---|
| `--phase <list>` | Comma list, or `all` / `off` (passed to `hailo_trace.phase=` via cmdline.txt) |
| `--mech <list>` | Same shape, for `hailo_trace.mech=` |
| `--shell-cmd "<cmd;…>"` | Optional post-boot commands; runs each via `labctl serial send` and appends output |
| `--scenario <name>` | Tag in the output filename |
| `--board <name>` | labctl SBC name (default `pi-5-1`) |
| `--skip-build` | Reuse `build/kernel/slmos.bin` |
| `--keep-cmdline` | Leave `cmdline.txt` on the card after capture (default: removed) |
| `--duration <minutes>` | Claim duration |
| `--out-dir <path>` | Override archive location |

## Idempotency

EXIT/INT/TERM trap removes `cmdline.txt` from the SD card and releases the labctl claim regardless of where the script fails. Re-running is always safe — no assumed prior card state.

## Test plan

- [x] `bash -n` syntax check passes
- [x] `shellcheck` 0.9.0 clean
- [x] `--help` / `-h` prints header, exit 0
- [x] No-args fails with clear message, exit 1
- [x] End-to-end run on `pi-5-1`: `scripts/capture-hailo-trace.sh --phase all --mech pci,rpc,irq --shell-cmd "hailo probe; hailo boot" --scenario script-validation --duration 15` produces `slmos-script-validation-pi-5-1-2026-05-10.txt` (280 lines, 11 KB). Better Phase 1 coverage than the manual PR A bootphase capture because cmdline.txt-armed masks are active from t=0 (caught DEVCTL/Command/LNKSTA reads at `off=0x88/0x90/0x92` that the manual run missed).
- [x] Cleanup trap verified: cmdline.txt removed, claim released; pi-5-1 free after run.

### DIFF
--- a/scripts/capture-hailo-trace.sh
+++ b/scripts/capture-hailo-trace.sh
@@ -1,0 +1,226 @@
+#!/usr/bin/env bash
+# capture-hailo-trace.sh — orchestrate a Hailo boundary-trace capture on
+# a Pi 5 lab board.
+#
+# Builds an SLM-OS kernel with the requested trace masks armed via
+# cmdline.txt, deploys to a labctl-managed SBC, captures serial output
+# through boot + (optionally) a follow-on shell command, and saves the
+# capture to ~/slmos-ref/derivatives/slmos-traces/.
+#
+# Dependencies (runtime):
+#   - PR A (feat/hailo-trace-framework) merged: provides the
+#     hailo_trace_cmdline_parse() path and the `hailo trace` shell
+#     subcommand this script depends on.
+#   - labctl CLI in PATH.
+#   - aarch64-none-elf-gcc toolchain.
+#
+# Usage:
+#   scripts/capture-hailo-trace.sh \
+#       --phase link,fw_boot,postboot \
+#       --mech  pci,rpc,irq \
+#       --shell-cmd "hailo probe; hailo boot" \
+#       --scenario bootphase \
+#       [--board pi-5-1] [--skip-build] [--duration 30] [--keep-cmdline]
+#
+# Exit codes:
+#   0  success — capture saved at the printed path
+#   1  argument / preflight error
+#   2  build failed
+#   3  labctl claim / sdwire / serial step failed
+#
+# Idempotency:
+#   On any failure after the SBC is claimed, the trap restores the SD
+#   card to a clean state (removes cmdline.txt) and releases the
+#   claim. A claimed-board interrupt (Ctrl-C) is treated the same way.
+#   Re-running the script is always safe — it does not assume any
+#   prior cmdline.txt state on the card.
+
+set -euo pipefail
+
+# ---- defaults --------------------------------------------------------------
+
+BOARD=pi-5-1
+SCENARIO=bootphase
+PHASE=""
+MECH=""
+SHELL_CMD=""
+SKIP_BUILD=0
+KEEP_CMDLINE=0
+DURATION_MIN=30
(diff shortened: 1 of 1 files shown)


johnjezl:
Pushed \`4ba46179\` addressing the actionable review findings.

**Warning — RETURN trap bug:**
- The \`trap '...' RETURN\` at the top of the "write cmdline.txt" section never fired (\`RETURN\` is function-scoped only). Tempfile leaked on any failure between \`mktemp\` and the explicit \`rm\`. Fixed by tracking \`\$TMP_CMDLINE\` as a file-scope variable that the main EXIT/INT/TERM \`cleanup()\` handler removes when non-empty, and clearing it on the success path so cleanup() no-ops.

**Suggestions:**
- **S1 — arg validation:** Added a \`require_val()\` helper called from every value-taking case arm. \`--phase\` with no value now prints \`error: --phase requires a value\` and exits 1 instead of aborting under \`set -u\`.
- **S2 — \`usage()\` hardcoded line range:** Switched from \`sed -n '2,30p'\` to an awk that prints every leading \`#\` line until the first non-\`#\` line. Verified: \`--help\` now outputs 35 lines (full header including the idempotency note, vs. 29 before).
- **S3 — labctl \`--until\` regex semantics:** Added a comment confirming labctl interprets \`--until\` as POSIX extended regex (so \`slmos> \$\` matches end-of-line, not a literal \`\$\`). Cross-references PR #781 / #783 hardware validation.
- **S5 — \`IFS=';'\` quoted-semicolon limitation:** Added a comment documenting the limitation and a possible fix path (switch to repeated \`--shell-cmd\` flag) for the day it actually bites someone.

**Skipped per review** ("not worth fixing"):
- **S4 — \`KERNEL_BIN\` hardcoded path.** Reviewer flagged "minor — the path has been stable for a long time"; agreed, leaving as-is.

**Bonus catch:** shellcheck SC2015 (\`A && B || C\` footgun) fired on my initial cleanup-fix attempt; rewritten as proper \`if/then\` to silence it.

Validation:
- \`bash -n\`: clean
- \`shellcheck\`: no findings (was clean before; still clean)
- \`--help\` line count: 35 (was 29)
- Hardware end-to-end revalidation deferred until next lab session — none of the changes affect the runtime flow (just arg parsing, comments, tempfile bookkeeping in the failure path that wasn't exercised in the original validation run).

Issue #784, opened 2026-05-11
CI workflow blocked: GitHub Actions org-level billing refusal since April 12


johnjezl:
## Symptom

The `Build and Test` workflow (`.github/workflows/ci.yml`) has not produced a clean run since **2026-04-12**. Every run since then either:

- Fails in 2-5 seconds with the annotation: *"The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings"*
- Is cancelled before completion (older April 12 runs)

`gh pr view 780 --json statusCheckRollup` returns `[]` — no CI is firing on PRs at all.

## Root cause

**Org-level GitHub Actions block on the SLM-OS organization**, not a workflow issue. The repo is public, but Actions can still be refused at the queue level when an organization's billing has a failed payment or spending-limit issue.

Verified by inspecting failed runs:
- Run 24319083111 (PR-triggered): same annotation across all 3 jobs.
- Run 24316777159 (push to main): same annotation across all 3 jobs.

The workflow file itself was *not* the blocker — jobs were refused before any step executed (the 2-5s runtime is just GitHub Actions queueing them and rejecting them).

## Fix required

Resolve the billing block at https://github.com/organizations/SLM-OS/settings/billing — update the failed payment method and/or raise the spending limit. This is account-administrative work; cannot be done via code change or `gh` CLI.

## Related workflow cleanup

The workflow file had accumulated significant drift since the last green run in April (new build flags, build-dir layout, QEMU safeguards) that would have caused the first post-billing run to fail anyway. Cleaned up in **#785** (`fix/ci-workflow-drift`) — switches to canonical `make test` targets so future Makefile changes don't require workflow edits.

## Acceptance

- [ ] Billing block resolved at the org level
- [ ] #785 merged
- [ ] Verify a `workflow_dispatch` run from a topic branch completes successfully (ARM job)
- [ ] x86-64 job continues to fail due to pre-existing rustc/LLVM ICE in slm-runtime (separate issue) — `continue-on-error: true` in #785 keeps the overall workflow green until that's fixed

johnjezl:
Closed by decision: SLM-OS will not enable GitHub Actions billing for this org. The workflow file at `.github/workflows/ci.yml` (cleaned up in #785) remains in-tree but dormant. Local `make test` and `make kernel PLATFORM=...` continue to be the canonical verification path; the workflow can be re-activated later by resolving the billing block if priorities change.

[...]

PR #785, opened 2026-05-11
ci: align workflow with current Makefile targets

johnjezl:
## Summary

Cleans up `.github/workflows/ci.yml` drift accumulated since the last green run on April 12. **Does not fix the underlying CI block** — that's org-level Actions billing (tracked in #784). This is the workflow-side preparation so the first run after billing is resolved is actually green.

## Changes

- **Switch to canonical `make test` / `make test PLATFORM=X86_64` targets.** The previous workflow invoked `cmake -B build ...` + `qemu-system-aarch64 ...` directly, which bypassed the Makefile's build-signature tracking, QEMU safeguards (cgroup memory + CPU caps via `systemd-run`, with graceful fallback when unavailable), and built-in PASS/FAIL detection. Now CI calls the same entry points local developers use, so future Makefile changes don't silently break CI.
- **Update artifact paths** to match the Makefile's actual output locations (`build/kernel-test/slmos.{elf,bin}`, `build/test-output.log`) — old paths pointed at `build/` directly.
- **Drop unused `gcc-aarch64-linux-gnu` apt install** — bare-metal kernel uses the separately-downloaded `aarch64-none-elf` toolchain, not the Linux-targeting one.
- **Add `grub-pc-bin`** to x86-64 deps (grub-mkrescue needs it alongside `grub-efi-amd64-bin`).
- **Add `concurrency.cancel-in-progress: true`** — new pushes cancel stale runs.
- **Add per-job `timeout-minutes`** — hangs surface as timeouts instead of burning the 6h default.
- **Mark x86-64 job `continue-on-error: true`** — pre-existing rustc/LLVM ICE in slm-runtime (*"Do not know how to soften this operator's operand"*) blocks the x86-64 build. Job still runs so we see when it starts passing; doesn't gate overall workflow status. Drop the flag once the ICE is fixed.

Net diff: 76 insertions, 122 deletions.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — valid YAML.
- [x] `make test` passes locally (verified earlier this session — full suite green including the 22 hailo_trace cases from #780).
- [ ] Run via `workflow_dispatch` after #784 is resolved to verify ARM job actually completes on a GitHub-hosted runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,100 +7,123 @@ on:
     branches: [ main ]
   workflow_dispatch:  # Allow manual trigger
 
+# Cancel in-progress runs for the same PR / branch on new pushes.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-and-test:
+    name: ARM64 build + test (QEMU_VIRT)
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
 
-    - name: Install ARM GCC Toolchain
+    - name: Install bare-metal ARM toolchain (aarch64-none-elf)
       run: |
-        sudo apt-get update
-        sudo apt-get install -y gcc-aarch64-linux-gnu
-        # Install the bare-metal toolchain
         wget -q https://developer.arm.com/-/media/Files/downloads/gnu/13.2.rel1/binrel/arm-gnu-toolchain-13.2.rel1-x86_64-aarch64-none-elf.tar.xz
         sudo tar -xf arm-gnu-toolchain-13.2.rel1-x86_64-aarch64-none-elf.tar.xz -C /opt
         echo "/opt/arm-gnu-toolchain-13.2.Rel1-x86_64-aarch64-none-elf/bin" >> $GITHUB_PATH
 
-    - name: Install QEMU
+    - name: Install QEMU + build deps
       run: |
-        sudo apt-get install -y qemu-system-arm
+        sudo apt-get update
+        sudo apt-get install -y qemu-system-arm cmake make
 
-    - name: Install Rust
+    - name: Install Rust (aarch64-unknown-none)
       uses: dtolnay/rust-toolchain@stable
       with:
         targets: aarch64-unknown-none
 
-    - name: Verify toolchain installation
+    - name: Verify toolchain
       run: |
-        aarch64-none-elf-gcc --version
-        qemu-system-aarch64 --version
+        aarch64-none-elf-gcc --version | head -1
+        qemu-system-aarch64 --version | head -1
         rustc --version
-        cargo --version
-        rustup target list --installed
+        cmake --version | head -1
 
-    - name: Build Runtime (Rust)
-      working-directory: runtime
-      run: cargo build
+    # `make test` is the canonical entry point — it drives the Makefile's
+    # build signature, QEMU safeguards (systemd-run cgroup limits when
+    # available, or direct invocation otherwise — see Makefile QEMU_GUARD),
+    # and PASS/FAIL detection via the kernel's isa-debug-exit + grep on
+    # crash markers. Direct cmake + qemu invocations bypass all of that
+    # and accumulate drift the moment the Makefile gains new build flags
+    # (which it has, often: HAILO_TRACE_*, JETSON_HW_TICK,
+    # SECONDARY_PREEMPT, WORK_STEALING, AI_SCHED, etc.).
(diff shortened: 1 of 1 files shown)


johnjezl:
Pushed `35653636` addressing the actionable review findings.

**Warnings:**
- `paths-ignore` (`docs/**`, `**/*.md`, `LICENSE`) added to both `push` and `pull_request` triggers — doc-only PRs no longer burn Actions minutes on the full matrix.
- The x86-64 job's `continue-on-error: true` comment now references the tracking issue (**#786** — `x86-64 build: rustc/LLVM ICE in slm-runtime`) so a future maintainer knows when to revert it.

**Suggestions:**
- ARM toolchain (~140 MB) now cached via `actions/cache@v4` keyed on toolchain version + `runner.os`. On cache hit the install step is skipped; PATH export factored to its own step so it runs in both branches.
- `concurrency.cancel-in-progress` scoped to `pull_request` events only — `main`/`develop` push runs now complete so rapid merges produce verifiable history per commit.
- Added workflow-level `permissions: contents: read` (minimum default).
- Artifact upload `if-no-files-found` flipped from `ignore` to `warn` so missing artifacts after a passing test step show up in the log.

**Not addressed** (review-explicit "out of scope" or "skip for now"):
- SHA-pinning third-party actions — separate hardening pass.
- Composite action for `apt-get update` — premature; only two jobs.

Filed **#786** to track the x86-64 ICE so the `continue-on-error` revert has a clear owner.

YAML still validates. Real verification still gated on #784 (billing).

johnjezl:
Pushed \`d828abe8\` addressing the actionable suggestions from the second review pass.

**Suggestions:**
- **S1 — toolchain version hardcoded in 3+ places:** Lifted \`ARM_TOOLCHAIN_VER\` and \`ARM_TOOLCHAIN_DIR\` to a workflow-level \`env:\` block. ARM's mixed-case URL/directory convention (\`rel1\` vs \`Rel1\`) makes the two-variable shape load-bearing — comment in the env block documents that.
- **S2 — \`paths-ignore\` doesn't cover other doc files:** Verified empirically: \`ls *.md *.txt CHANGELOG* AUTHORS* CONTRIBUTING* README*\` at top level returns only \`.md\` files (all covered by \`**/*.md\`), plus \`CMakeLists.txt\` and \`version.txt\` which are build inputs, NOT docs. Current globs are correct. No change.
- **S3 — tarball left in workspace:** Added \`rm -f "\$TARBALL"\` after \`tar -xf\`. Defensive against a future artifact glob accidentally including the 140 MB download.
- **S4 — \`Verify toolchain\` doesn't fail loudly:** Added \`set -e\` to both verify steps (ARM job + x86-64 job). A missing tool now fails the step immediately instead of being masked by \`head -1\`'s exit-0 on empty input.

**Warning:**
- **Cross-platform coverage gap (Pi 5 + Jetson):** Filed as #790. Pre-existing gap, not introduced by this PR; review marked it "out of scope for 'make the first post-billing run green' — file as follow-up, don't block." Done.

YAML still validates. Net total for the PR is now +103 / -123 → +122 / -128 across 3 commits.

johnjezl:
Pushed \`5b05613a\` addressing both remaining actionable findings.

**S1 — toolchain integrity verification:** Caveat: the review presumed SHA-256, but ARM only publishes MD5 for these tarballs (at the \`.tar.xz.asc\` URL — despite the \`.asc\` extension, the content is a plain MD5 hash, not a GPG signature). Implemented what's actually available:
- Pinned \`ARM_TOOLCHAIN_MD5: b4aa2d1c2d9c857e194e28a2a8271321\` in the env block.
- Added \`md5sum -c -\` after \`wget\` in the install step.
- Comment is explicit that MD5 is corruption detection, NOT supply-chain hardening (MD5 collisions are computable). Strictly better than no check — narrows the attack surface to "compromise both the git repo and the ARM CDN."
- Stronger verification (TOFU-pinned SHA-256 we compute ourselves, or GPG if ARM ever starts signing) is left as a future hardening pass if threat model warrants.

**S2 — Actions expression vs shell expansion syntax:** Added a "Syntax note" paragraph to the env-block comment explaining that step-level fields use \`\${{ env.NAME }}\` (Actions evaluates before shell starts) while \`run:\` bodies use \`\${NAME}\` (workflow env auto-exported as shell env). Mixing or normalizing to one form will break the cache step. Documentary only.

**S3 — \`set -e\` redundancy:** Review marked "No change requested." Kept as-is.

This should close the loop on actionable findings for PR #785. All three review passes' substantive items are now addressed or deferred with explicit follow-up issues:
- #786 — x86-64 rustc/LLVM ICE (deferred via \`continue-on-error: true\`)
- #790 — Pi 5 / Jetson CI compile coverage (filed as follow-up)

Three commits added; total now +148 / -130 net across 4 commits on the PR.

Issue #788, opened 2026-05-11
GA10B dispatcher wedge during Qwen-scale W3 staging: FECS_ERROR_PENDING


johnjezl:
## Summary

Reproduced the dispatcher wedge from #779 (which had been closed as not-reproducing on SmolLM2) under a Qwen-scale W3 weights-staging workload. The diagnostic shipped in PR #782 caught the actual fault class on the first attempt.

## Diagnostic capture

During Qwen-2.5-1.5B-Instruct-Q4_K_M xload + W3 staging on jetson-nano-1 (2026-05-11):

\`\`\`
[ce-memcpy] GP_GET did not advance — PBDMA didn't see our submit
[ce-memcpy]   gr_intr=0x00080000 gr_exception=0x00000000 class_error=0x00000000 trapped_addr=0x80000000 fe_hww_esr=0x80000000 fecs_intr=0x00000000
[ce-memcpy] GP_GET did not advance — PBDMA didn't see our submit
[ce-memcpy]   gr_intr=0x00080000 ...
[ce-memcpy] GP_GET did not advance — PBDMA didn't see our submit
[ce-memcpy]   gr_intr=0x00080000 ...
\`\`\`

Decoding (per \`~/slmos-ref/nvidia/nvgpu-hw-gr-ga10b.h\`):
- \`gr_intr=0x00080000\` → bit 19 = \`gr_intr_fecs_error_pending_f\` (line 74). **The FECS microcontroller errored out.**
- \`trapped_addr=0x80000000\` → bit 31 (\`VALID\`) set; subchannel = 0, method = 0 (SET_OBJECT)
- \`fe_hww_esr=0x80000000\` → bit 31 (\`VALID\`) set, real FE hardware-wedge exception

Once FECS is in error, PBDMA refuses to advance the channel; every subsequent CE submit times out the same way (2 s × thousands of remaining chunks = hours of stuck wall time).

## Repro

1. \`gpu-channel-helper --preserve-for-kexec --qmd-pool --weights-pool-size 256M\` (smaller pool keeps PMM headroom — see follow-up #TBD-A for the 1.5 GB case)
2. \`slmos-kexec --no-helper\`
[...]

PR #801, opened 2026-05-12
mm: pmm — add user-reserve API + Jetson kexec inst-block reservation (#788 Stage 1)

johnjezl:
## Summary

Adds infrastructure for runtime-discovered memory reservations that SLM-OS's PMM must honor before publishing pages to the buddy allocator. Primary use case: the GPU kexec handoff state that `gpu-channel-helper`'s nvgpu allocated pre-kexec, which SLM-OS would otherwise clobber with its own kernel allocations (root cause of the #788 wedge — proven by `nvgpu instdump` in PR #797).

**This is Stage 1 of a multi-stage fix.** It adds the reservation scaffolding and demonstrates protection of one critical page (the channel inst block). It does NOT by itself eliminate the wedge — Linux's kexec also actively frees kernel-internal nvgpu allocations, so a single-page reservation isn't sufficient. The scaffolding is reusable by whichever Stage 2 path lands.

## New PMM API

```c
int pmm_user_reserve_add(uint64_t phys, uint64_t size);
size_t pmm_user_reserve_count(void);  /* test-only */
void   pmm_user_reserve_reset(void);  /* test-only */
```

`pmm_user_reserve_add` registers a range that PMM init carves out alongside the firmware-supplied `/memreserve/` entries from the DTB. Both lists feed the existing `pmm_carve_reserves` helper, so the buddy never sees the reserved pages. Capacity is 16 ranges (`PMM_MAX_USER_RESERVES`), generous for the kexec-handoff caller.

Must be called before `pmm_init`; late registrations are silently ignored (documented in the API).

## Jetson hook

`ga10b_kexec_handoff_register_reserves` (new `kernel/gpu/nvidia/ga10b_handoff_reserve.c`) reads `FECS_CURRENT_CTX` via the BAR0 identity mapping `vmm_init` just established, decodes the channel inst-block phys, validates it sits inside DRAM via `ga10b_phys_in_dram`, and registers a single 4 KB reserve.

Called from `kernel_main` between `vmm_init` and `pmm_init` — the only window where the MMU is up but PMM hasn't yet built free lists. Safe on non-kexec boots: early-returns when `FECS_CURRENT_CTX` reads zero / target=0 / a `0xbadfXXXX` poisoned pattern (any of which indicate "no preserved channel").

## Hardware verification on jetson-nano-1 (2026-05-12)

`[ga10b-reserve] reserved kexec'd channel inst block at phys=0x122c67000` prints during boot before `pmm_init`. After kexec the reserved page is no longer clobbered with SLM-OS kernel code or model bytes (the previous failure mode from PR #797 `nvgpu instdump`).

The page now reads as **all zeros** instead of garbage. Better, but still not a valid nvgpu inst block.

## Known limitation: the wedge still fires

`slm xload qwen` still wedges on chunk 0 of W3 staging after this fix, with a slightly different fault pattern (`fault_info.valid=0` instead of the previous `INVALID_PDE`).

**Why:** USERSPACE-allocated dmabufs (USERD, GPFIFO, PB, SEM, SASS pool, weights pool) survive kexec intact — verified via `peek 0x113b25000 64` showing the USERD page with `GP_PUT=1, GP_GET=1` and USERD entry fields populated as the helper left them. But the kernel-internal nvgpu allocations (inst block, page tables, GR ctxsw save buffer) appear to be freed by Linux's `module .shutdown` handler running on kexec — by the time SLM-OS reads `FECS_CURRENT_CTX`, the register still names the old phys but the page itself has been freed and zeroed.

The all-zeros inst block has zeros at words 0/1 (PDB pointer). When PBDMA loads the channel for the first CE memcpy, it walks PDB=0 → fault.

## Stage 2 paths (separate PRs)

1. **Patch nvgpu kernel-side** to skip channel teardown on kexec. Largest scope, requires Linux kernel rebuild.
2. **Rebuild GMMU state in SLM-OS post-kexec.** Allocate fresh inst block + PDB + page tables, rebind via FECS protocol. Tractable but multi-session.
3. **Open a fresh channel post-kexec via direct hardware ops** — bypass nvgpu's preserved channel entirely. Cleanest but largest blast radius.

Each Stage 2 option will use the `pmm_user_reserve_add` API landing here.

## Test plan

- [x] `make test` passes (QEMU ARM64; 5 new unit tests for the API: count-starts-zero, add-increments-count, zero-size-rejected, reset-zeros-count, table-full-rejects)
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO` builds clean
- [x] `make kernel PLATFORM=RASPI5` builds clean (non-Jetson stub of `ga10b_kexec_handoff_register_reserves` compiles cleanly; the function is platform-guarded at the call site too)
- [x] Hardware test on jetson-nano-1: reserve-message prints during boot; reserved page contents change from "ARM64 kernel code / Qwen weight bytes" → "all zeros"; USERD survival verified via `peek`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -486,6 +486,11 @@ set(KERNEL_SOURCES
     # channel; no CPU-side phys arithmetic required, sidesteps the
     # GMMU walk-discovery failure observed post-kexec.
     $<$<STREQUAL:${PLATFORM},JETSON_ORIN_NANO>:kernel/gpu/nvidia/ga10b_ce.c>
+    # Jetson: register the kexec'd channel's inst block as a PMM
+    # user-reserve before PMM init publishes pages, so nvgpu state
+    # surviving in DRAM doesn't get clobbered by SLM-OS kernel
+    # allocations. #788 root-cause fix.
+    $<$<STREQUAL:${PLATFORM},JETSON_ORIN_NANO>:kernel/gpu/nvidia/ga10b_handoff_reserve.c>
     # Jetson: full GSP platform shim — MMIO at GPU_BASE, unified memory
     # DMA, cache maintenance, firmware via .incbin. Replaces the old
     # nvidia_gsp_platform_stub.c now that CBB firewall is bypassed.
--- a/kernel/gpu/nvidia/ga10b_handoff_reserve.c
+++ b/kernel/gpu/nvidia/ga10b_handoff_reserve.c
@@ -1,0 +1,160 @@
+/*
+ * ga10b_handoff_reserve.c — implementation of
+ * `ga10b_kexec_handoff_register_reserves`. See the header for the
+ * #788 problem statement.
+ *
+ * The function reads FECS_CURRENT_CTX (BAR0 + 0x409b00), decodes the
+ * inst-block physical address per `gr_fecs_current_ctx_*`
+ * (bits[27:0] = inst_phys >> 12, bits[29:28] = aperture target),
+ * bounds-checks it against the DRAM aperture via `ga10b_phys_in_dram`,
+ * and registers a single 4 KB reserve via `pmm_user_reserve_add`.
+ *
+ * Why read FECS_CURRENT_CTX directly here instead of going through
+ * `ga10b_gmmu_discover_inst_block_phys`?
+ *
+ * The MMU is up at the call site (vmm_init has run, identity-maps
+ * cover all DRAM + the GPU BAR0 aperture), but the higher-level
+ * GPU bringup infrastructure (`ga10b_bringup` state machine, PMM-
+ * dependent allocations inside `ga10b_gmmu_*`) has not been
+ * initialised — `pmm_init` itself is the next call after us. Reading
+ * the raw MMIO directly keeps the dependency footprint to the
+ * identity mapping plus `pmm_user_reserve_add`, both of which are
+ * known-safe to use this early.
+ *
+ * **Limitations of the MVP:**
+ *
+ *   - Reserves only the inst block. The PDB and page-table pages
+ *     pointed to from the inst block also need protection but the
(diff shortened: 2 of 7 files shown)


johnjezl:
Closing as superseded.

The architectural premise behind the #788 stage series — "reserve PDB tree pages in PMM so SLM-OS doesn't clobber them post-kexec" — was **specifically disconfirmed** during the 2026-05-17 session investigating #834.

The "PDB tree pages are clobbered" conclusion came from the GMMU walker reporting valid VAs as unmapped. With the walker bit-position bug fixed in commit 5977ef40 (on branch `oplib-stage-rebuild-gmmu` / PR #845), the walker now resolves Linux's PDB end-to-end with leaf phys values matching the helper-published handoff exactly. The PDB tree was never being clobbered — we were reading the wrong table offsets because the walker's VA-bit constants didn't match nvgpu's canonical `ga10b_mm_levels[]`.

What the session-final state of the buffer-reservation story actually looks like:

- **a099cd3b** — `ga10b_validate_handoff` rejects `inst_block_phys=0` so stale handoffs left in DRAM by prior helper invocations don't mask fresh ones.
- **8b099aeb** — `ga10b_kexec_handoff_register_reserves` uses the kind-filtered scanner (instead of the bare magic-match) so the boot reserve pins the FRESH handoff's userd/gpfifo/pushbuf/sem + v7 qmd_pool + v10 gr_ctx extents. (This was the only "buffers were getting clobbered" problem that actually existed, and the helper had been publishing v10 gr_ctx extents all along — they were hidden behind the stale handoff.)

Neither of those needed a new `pmm_user_reserve` API — the existing `pmm_user_reserve_add` was already in tree.

If a future need for a more general PMM user-reserve API surfaces, this PR's design can be re-extracted from history. Closing because there's no remaining call site that needs the work as-implemented here.

Issue #840, opened 2026-05-15
ga10b_gmmu walker: dual PDE0 only follows small-page side, hides big-page mappings


johnjezl:
## Problem

The static GMMU walker at `kernel/gpu/nvidia/ga10b_gmmu.c:206` (`ga10b_gmmu_walk`) only follows the small-page subtree at PDE0 (the dual PDE). Big-page mappings are silently mis-classified as unmapped.

Header documents this as an intentional Milestone A limitation (`ga10b_gmmu.h:90-93`), under the assumption that handoff dmabufs are 4 KB-mapped. In practice, NVIDIA's nvgpu often picks 64 KB big pages for buffers ≥64 KB / 64 KB-aligned — the handoff's pushbuffer is exactly 64 KB. If Linux chose big-pages for any handoff buffer, the walker reports it as unmapped even when it's actually present in the big subtree.

Surfaced during #834 investigation 2026-05-14: walks of `pushbuf_gpu_va` against `h->inst_block_phys` show `PDE2[127] INVALID` (i.e. the failure is *before* PDE0, so this gap isn't what's failing here), but the walker output was misleading for hours because we couldn't distinguish "really unmapped" from "big-page mapping the walker can't see."

## Fix (already drafted on this branch)

Replace the small-only `decode_dual_pde_small` call in `ga10b_gmmu_walk` with a new `decode_dual_pde` helper that tries small first, falls back to big when small is invalid. The walker's per-level record already has a `dual_pde_followed_small` flag that gets set to `false` when the big subtree is taken — so the printed walk output disambiguates which subtree was descended.

The writer-path `ensure_dual_pde_small_table` keeps using the small-only decoder — SLM-OS-side mappings always author 4 KB PTEs, so writers only ever care about the small subtree.

Diff is in `kernel/gpu/nvidia/ga10b_gmmu.c` and `ga10b_gmmu.h` on the current working tree (not yet committed pending #834 progress).

## Why it's only P2

The dual-PDE walker gap doesn't cause #834's failure (which aborts at PDE2[127] before reaching PDE0). It IS a real bug for any case where Linux mapped a buffer via big-pages and we need to walk it from a static investigation. Fixing it now is cheap and prevents future investigations from going down the same misleading path.

## Acceptance

- Walker correctly identifies a 64 KB big-page mapping at PDE0 (write a hardware test or contrived synth test)
[...]

PR #841, opened 2026-05-15
gmmu: walker follows big-page side of dual PDE0 too (#840)

### LINKED ISSUE #834: W-series dispatch: ga10b_dispatch_v7_pipeline_inline returns -1 on every call (#832 follow-on)
## Symptom

After PR for #832 lands (oplib_pool self-allocates 256 KB scratch + 4 KB cbuf when the helper's v6/v7 handoff zeros `h->shader_*` and `h->cbuf_*`), the W-series dispatch FFI gets past its entry guards but every dispatch fails downstream with:

```
[oplib-dispatch] inline dispatch failed: rc=-1
[GA10B-P8-v7-DBG] ops_v7 array @0x13cf12000, n=1
[GA10B-P8-v7-DBG]   op[0] shader_gpu_va=0x4000000600 cbuf_gpu_va=0x4000040000 qmd_gpu_va=0x1ffc0c0000 output_phys=0x1ffc0c0000 reg_v=64 shader_phys=0x0 shader_size=0
```

The `[GA10B-INST-CTX]` diagnostic that fires on dispatch failure shows the inst-block's second half (offsets 0xe10..0xff0) is all zeros, and `[GA10B-INST-FAULT] fault_info.valid=0` — no GPU fault latched.

## What's working

- `slm_ffi_resolve_scratch_base()` correctly resolves to the pool's base addresses:
  - `shader_gpu_va = 0x4000000600` (pool base + 0x600, the RMSNORM SASS within the staged operator library)
  - `cbuf_gpu_va = 0x4000040000` (pool cbuf)
- The dispatch FFI memcpys input/gamma/output into the pool's scratch slots
- `slm_oplib_prepare_dispatch` constructs a v7 op with correct shader_gpu_va, cbuf_gpu_va, register_count_v
- `slm_oplib_dispatch` calls `ga10b_dispatch_v7_pipeline_inline(b, &op, 1)` with the single op

## What's failing

`ga10b_dispatch_v7_pipeline_inline` returns -1. This is the inner v7 dispatch loop that:

- Builds a pushbuffer entry for the op (METHOD_BIND_GROUP / SEND_PCAS2_B / etc.)
- Submits via the channel's GPFIFO + USERD
- Polls the trailing semaphore for completion

One of those steps is failing. The diagnostic dump shows post-failure state but doesn't pinpoint which step.

## Possible causes (in rough order of likelihood)

1. **QMD pool exhaustion or stale state.** The W-series FFI uses `h->qmd_pool_gpu_va` as the QMD placeholder (sentinel for "v7 op valid"). If the QMD pool is consumed by MNIST's pre-kexec pushed ops + never reset, new dispatches can't allocate a fresh QMD slot.
2. **Pushbuffer space exhaustion.** The MNIST helper's pre-kexec push consumed pushbuffer slots; SLM-OS's W-series dispatch tries to append but `gp_put`/`gp_get` boundary is wrong.
3. **Semaphore phys mismatch.** The dispatch waits on a semaphore but the placeholder phys (currently `h->qmd_pool_gpu_va`, which is a GPU VA, not a phys) is invalid.
4. **Channel state mismatch.** The W-series single-op dispatch on a channel that was set up for an 8-op MNIST pipeline may require explicit pipeline reset / state-machine teardown before single-op dispatches work.

## Investigation pointers

- `kernel/gpu/nvidia/ga10b_bringup.c` — `ga10b_dispatch_v7_pipeline_inline` source
- The `[GA10B-INST-CTX]` and `[GA10B-INST-FAULT]` diagnostic was added in #797; understanding which check actually fired and why would help
- PR #828's `[GA10B-P8-v7-DBG]` log shows the v7 op contents at submit time — useful to compare against a known-working MNIST dispatch
- Per the `op[0]` log line, `qmd_gpu_va = 0x1ffc0c0000` is the helper's qmd_pool_gpu_va. That GPU VA needs to be mapped in the channel (it is — it's a helper allocation) and have at least one free QMD slot.

## Hardware repro (verified 2026-05-14)

After PR #832 fix lands:

```bash
make kernel PLATFORM=JETSON_ORIN_NANO
scripts/jetson-deploy-retry.sh -k build/kernel/slmos.elf
python3 /tmp/slm_xload.py 192.168.4.5:2323 smollm <SmolLM2-135M.gguf>
```

From the shell:

```
slmos> gpu use inference on
slmos> nvgpu oplib stage
slmos> slm launch 0 --ctx 256 --sampler greedy
slmos> slm prompt 0 Hello
```

Observe `[oplib-dispatch] inline dispatch failed: rc=-1` per attempted dispatch.

## Acceptance

After fix:
- `slm gpu` shows `ok > 0` for the W-series transformer ops on the toggle-respecting path
- A diff between CPU-only and GPU-on `slm prompt` outputs shows at least byte-level differences in the FC-layer logits (`bd...` family of FP32 bits)

## Related

- #832 — the entry-guard bug this PR's fix unblocks
- #714 — W-series umbrella
- #764 — SLM-GPU-end-to-end rollup (this is the next blocker)
- PR for #832 fix

johnjezl:
## Summary

- `decode_dual_pde_small` in the GA10B GMMU walker only ever followed the small-page subtree at PDE0; big-page mappings were silently mis-classified as unmapped.
- Add `decode_dual_pde` that tries the small subtree first and falls back to big when small is invalid.
- Keep `decode_dual_pde_small` for the writer path (`ensure_dual_pde_small_table`), which only authors 4 KB PTEs.

Surfaced during the #834 investigation. Doesn't itself fix #834 — the failing walks abort at PDE2[127] before reaching PDE0 — but removes a misleading diagnostic that ate hours of investigation time.

## Test plan

- [x] `make kernel PLATFORM=JETSON_ORIN_NANO` builds clean
- [ ] Hardware: `nvgpu gmmu walk-raw <inst> <va>` on a buffer Linux mapped via 64 KB big pages should now succeed (walker output prints `dual: big` instead of `dual: small`)
- [ ] No regression on `nvgpu rebuild-gmmu` writer path (still uses small-only via `ensure_dual_pde_small_table`)

### DIFF
--- a/kernel/gpu/nvidia/ga10b_gmmu.c
+++ b/kernel/gpu/nvidia/ga10b_gmmu.c
@@ -171,8 +171,10 @@ static void decode_regular_pde(uint64_t entry,
     *out_next_phys = ((addr_hi << 24) | addr_lo) << 12;
 }
 
-/* Decode the small-page half of a dual PDE (PDE0). Big half is
- * ignored for Milestone A — see ga10b_gmmu.h. */
+/* Decode the small-page half of a dual PDE (PDE0). Used by the
+ * writer-side `ensure_dual_pde_small_table` helper, which only ever
+ * needs the small-page subtree because SLM-OS-side mappings author
+ * 4 KB PTEs. */
 static void decode_dual_pde_small(uint64_t entry,
                                   uint64_t *out_next_phys,
                                   uint8_t *out_aperture,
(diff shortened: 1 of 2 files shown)


johnjezl:
Closing as superseded by commits on branch `oplib-stage-rebuild-gmmu` (PR #845):

- **5977ef40** — fixes the walker's underlying VA-bit-position bug. The constants `GA10B_PDE3_VA_HI` through `GA10B_PTE_VA_LO` in `ga10b_gmmu.h` were a 49-bit-VA placeholder that never got tuned against hardware (the source comment literally admitted this). The corrected positions match nvgpu's canonical `ga10b_mm_levels[]` in `nvgpu-l4t-r36.4.4-gmmu_ga10b_fusa.c:353`.
- **9eb76083** — adds dual-PDE0 handling (big-page child first at offset 0..7, small-page fallback at 8..15) PLUS a per-child PTE bit-width override (big-page = bits[20:16] = 5 bits; small-page = bits[20:12] = 9 bits).

#841's small-first / big-fallback choice partially addresses the same surface issue but keeps the old wrong VA-bit constants — so it wouldn't have actually made the walker reliable. The deeper bit-position bug was what caused "VA unmapped" reports for VAs that PBDMA was happily translating in the same boot.

Hardware-verified on jetson-nano-2 2026-05-17: walker resolves both pushbuf VA (0x1ffc000000 → 0x13be40000) and sema VA (0x1ffc08b000 → 0x13ba74000) end-to-end through Linux's PDB, both leaf phys values matching the helper-published handoff exactly.

Reopen if there's something in this PR's approach that the session's commits don't cover.

Issue #855, opened 2026-05-16
XGBoost sched policy 58c: SCHED_MODEL_KIND_XGBOOST + sched_xgb.c + derived features


johnjezl:
## Description

Sub-ticket of #58. Wire the trained XGBoost cascade from #58a (#850) through the generalized traversal engine from #58b (#854) into a new pluggable scheduler policy registered alongside `ai_mlp`, `ai_ppo`, and `ai_hailo`.

### 1. Scheduler blob kind (sixth slot)

Add `SCHED_MODEL_KIND_XGBOOST = 6` to the scheduler blob kind enum. Touch points (greps already done):
- `kernel/include/sched_model.h` (or wherever the existing kind ids live)
- `kernel/src/lua_slm.c:2402` — `sched_model_kind_id()` name→id map
- `kernel/src/shell_sys.c:6421` — `sched_model_kind_id()` shell-side map
- `kernel/src/shell_sys.c:6431` — `sched_model_kind_name()` reverse map
- `kernel/src/blob_autoload.c` — autoload routing (existing `sched_blob_validate_file` / `sched_blob_stage_file` paths)
- `kernel/src/runtime_blob_file.c:190` — `sched_blob_stage_file` accepts the new kind

### 2. Rust predictor

New file `runtime/src/sched/xgb.rs`:
- Parses staged blob via the generalized `parse_payload` from #58b (cascade form, N=3).
- Stores three classifier graphs.
- Exposes `predict(state: &[f32; 108], derived: &[f32; 5]) -> (u8 core, u8 priority, u8 preempt)`:
  - Build per-classifier feature slice. `core_clf` consumes 113 features. `priority_clf` consumes 114 (113 + `core` as extra). `preempt_clf` consumes 115 (114 + `priority`). Match the Python cascade ordering exactly — see `~/projects/slm-os-scheduler-ai/training/xgboost/train.py:60 TripleClassifier.predict`.
  - Argmax over tree sums per classifier; map through `label_classes` from the blob's per-classifier label map.
- Round-trip test: load the `test_vectors_xgb.bin` from #58a, run predict on each, compare against `expected_actions_xgb.bin`. Must match bit-for-bit.

### 3. Derived features

[...]

Issue #854, opened 2026-05-16
XGBoost sched policy 58b: generalize runtime_xgboost.rs traversal engine


johnjezl:
## Description

Sub-ticket of #58. The eviction-side XGBoost tree-traversal engine (`runtime/src/mm/eviction/runtime_xgboost.rs`, 281 LOC) is currently shape-locked to `BlockFeatures`. Generalize it so the scheduler can reuse the same parser, the same node-walking loop, and the same bounds-checked / cycle-safe invariants, but with its own feature shape (108-dim state + 5 derived features = 113 floats).

## Scope

1. Move shared logic to a new module — proposed `runtime/src/ml/xgb_tree.rs` (or wherever fits the crate layout). Parameterize the traversal on a feature-slice type:
   - Either a trait `XgbFeatures { fn get(&self, idx: usize) -> f32; }` with implementations for `BlockFeatures` (eviction) and `&[f32; 113]` (sched).
   - Or a plain `&[f32]` slice argument — simpler, more permissive, callers handle their own shape constants.
2. Keep all existing safety properties: bounds-checked child indices, finite-value asserts, cycle-protection iteration budget (16K node ceiling), `RuntimeXGBoostError::*` variants.
3. Make the parser accept multi-classifier payloads (the sched cascade is three classifiers in one blob; eviction is a single classifier). Extend `parse_payload` to handle a cascade header that names N sub-classifiers; eviction continues to use N=1.
4. Refactor `runtime/src/mm/eviction/xgboost.rs` and `runtime/src/mm/eviction/runtime_xgboost.rs` to call into the shared module. Regression-test parity: identical predictions for identical inputs against the pre-refactor code path (the existing eviction policy ABI must not change).

## Acceptance criteria

- `cargo test` in the runtime crate passes (existing eviction tests + new shared-module tests).
- Eviction `bench eviction-policy` numbers unchanged within noise (~1% latency tolerance).
- New module compiles standalone (no eviction-side dependencies leak into it).
- `parse_payload` accepts both single-classifier (eviction, N=1) and cascade (sched, N=3) shapes.

## Out of scope

- Sched-side blob kind registration and FFI → #58c (#852).
- New trained-model formats → handled by #58a (#850) in the sibling repo.

## Estimated Effort

~1-2 days.

## References

- Parent: #58. Umbrella: #449.
[...]

Issue #58, opened 2026-04-12
XGBoost scheduler policy (optional, runtime-loaded)


johnjezl:
## Description

Add XGBoost (gradient-boosted decision tree) as an **optional swappable scheduler policy** alongside the existing `heuristic`, `ai_mlp`, `ai_ppo`, and `ai_hailo` policies. SLM-OS is an exploratory operating system in which pluggable policies are a first-class design value — see #848. A new policy's purpose is to expand the option space available to researchers; it does NOT need to beat existing policies on benchmarks to be worth shipping.

**Runtime-loaded only.** Weights are staged/activated/rolled-back via the existing scheduler blob subsystem (`slm.sched_model_status / activate / rollback / clear`, `sched_blob_stage_file`, `blob_autoload`). No in-kernel-image embedding — the cascade is ~6-13 MB depending on serialization, which would balloon the image vs. ~515 KB for MLP/PPO weights. Runtime loading also mirrors how XGBoost already ships for the eviction subsystem.

## What's already in place (do not redo)

**In `~/projects/slm-os-scheduler-ai` (sibling repo):**
- Three classifiers trained: `models/xgboost/{core,priority,preempt}_clf.json` (27 MB + 552 KB + 440 KB).
- Tree-flattening + C-source emitter: `scripts/export_models.py::_flatten_tree`, `write_xgboost_c`. The C-source form was rejected by Plan A; the underlying flatten logic + 14-byte `xgb_node` record format are reusable for binary output.
- Three-classifier cascade prediction logic in Python: `training/xgboost/train.py::TripleClassifier.predict`. Cascade ordering: `core_clf → priority_clf(features + core) → preempt_clf(features + core + priority)`.
- 5 derived features defined in Python: `training/xgboost/features.py` (`max_deadline_urgency`, `core_util_std`, `deadline_task_count`, `gpu_should_use`, `best_cache_fit_core`).
- Full head-to-head evaluation vs MLP / PPO / experts: `results/eval_summary.csv` (200 episodes × 9 scenarios).

**In this repo:**
- Scheduler policy vtable: `struct sched_policy_ops` in `kernel/include/sched_policy.h`.
- 108-dim state extractor: `ai_extract_state()` in `kernel/sched/ai/ai_state.c`.
[...]

PR #883, opened 2026-05-16
runtime/ml: shared XGBoost tree-traversal engine (#854)

johnjezl:
## Summary

Sub-ticket [#854](https://github.com/SLM-OS/SLM-Operating-System/issues/854)
(part of XGBoost scheduler policy umbrella
[#58](https://github.com/SLM-OS/SLM-Operating-System/issues/58)).

Moves the eviction-side XGBoost parser + tree walker into a new
`runtime/src/ml/xgb_tree.rs` so the upcoming AI scheduler cascade
policy ([#855](https://github.com/SLM-OS/SLM-Operating-System/issues/855))
can reuse the same bounds-checked, cycle-safe loader without
copying 280 LOC.

- **XGB1** (16 B/node, u16 children, ≤ 65 535 nodes) — single
  classifier, used by the eviction `XGBoostPolicy` unchanged.
  `runtime_xgboost.rs` becomes a thin wrapper preserving every
  public symbol the eviction module + `lib::rust_eviction_run_tests`
  depend on.
- **XGBC** (20 B/node, u32 children) — N-classifier cascade for
  the AI scheduler. Widened indices required: trained `core_clf`
  flattens to ~450 K nodes, well past the u16 ceiling. Parser takes
  a per-classifier `max_feature_idx` slice so each sub-classifier
  can be pinned to its own input width (sched cascade is 113 / 114
  / 115 features per classifier).

Eviction predictions are unchanged — `predict_sigmoid` is bit-for-bit
the same code as before. New `predict_argmax` / `predict_label`
helpers exist to support the multiclass cascade in #855 and have
no eviction-path caller.

Safety properties preserved across the refactor: child indices
bounds-checked at parse against per-classifier `node_count`,
non-finite thresholds + leaf values rejected, tree-walk depth
capped at `MAX_TREE_DEPTH = 256` to detect cycles in
malformed-but-checksum-valid blobs (PR #465 regression — covered
by the `eval_tree_breaks_cycle_with_fallback` test, ported into
the new module).

## Dependency

Should merge before [#855](https://github.com/SLM-OS/SLM-Operating-System/issues/855)
(which calls into `crate::ml::xgb_tree::parse_cascade`).

## Test plan

- [x] `make test` (QEMU) — green; eviction `XGBoost*` smoke tests
  unchanged and passing
- [x] `cargo build --features ai_eviction` clean for both
  `aarch64-unknown-none` and `x86_64-unknown-none`
- [x] `make kernel PLATFORM=RASPI5 AI_SCHED=ON` — clean
- [ ] `make kernel PLATFORM=JETSON_ORIN_NANO AI_SCHED=ON` —
  pre-existing fails the linker "Kernel image too large" assertion
  in `kernel/kernel-jetson.ld` (also fails on origin/main without
  this PR; out of scope)

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -26,6 +26,7 @@ use linked_list_allocator::LockedHeap;
 
 pub mod kernel_ffi;
 pub mod log;
+pub mod ml;
 pub mod mm;
 pub mod sched;
 pub mod component;
--- a/runtime/src/ml/mod.rs
+++ b/runtime/src/ml/mod.rs
@@ -1,0 +1,8 @@
+//! Generic ML primitives shared across subsystems.
+//!
+//! Currently hosts the XGBoost tree-traversal engine (`xgb_tree`) used
+//! by both the eviction `XGBoostPolicy` and (per #855) the AI scheduler
+//! cascade policy. New shared inference helpers (e.g. small MLPs,
+//! adaptive predictors) belong here.
+
+pub mod xgb_tree;
--- a/runtime/src/ml/xgb_tree.rs
+++ b/runtime/src/ml/xgb_tree.rs
@@ -163,9 +163,20 @@ pub fn parse_single(bytes: &[u8], max_feature_idx: usize) -> Result<XgbModel, Xg
         return Err(XgbError::TooManyNodes);
     }
 
-    let roots_len = tree_count * 2;
-    let nodes_len = node_count * NODE_LEN_V1;
-    let expected_len = SINGLE_HEADER_LEN + roots_len + nodes_len;
+    // `MAX_TREES_SINGLE * 2` and `MAX_NODES_SINGLE * 16` both fit
+    // comfortably in usize on every supported platform, but use
+    // `checked_*` for consistency with `parse_classifier_section`
+    // and to guard against a future cap bump.
+    let roots_len = tree_count
+        .checked_mul(2)
+        .ok_or(XgbError::BadLength)?;
+    let nodes_len = node_count
+        .checked_mul(NODE_LEN_V1)
+        .ok_or(XgbError::BadLength)?;
+    let expected_len = SINGLE_HEADER_LEN
+        .checked_add(roots_len)
+        .and_then(|n| n.checked_add(nodes_len))
+        .ok_or(XgbError::BadLength)?;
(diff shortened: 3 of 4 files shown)


johnjezl:
slmos-review pass 2 — addressed in 8fb83b6f:

- `predict_argmax` now stack-allocates the per-class score buffer (`[f32; MAX_LABEL_CLASSES]`) instead of `vec![0.0; n_classes]`. Cascade callers in #885/#886 hit this 3× per scheduling decision (one per classifier) from IRQ context on hardware-tick paths, so the per-call heap alloc was the wrong default. Defence-in-depth clamp added in case a non-`parse_cascade` caller hands in `n_classes > 64`.
- `parse_single` length math switched to `checked_mul`/`checked_add` for consistency with `parse_classifier_section`. Caps fit in `usize` today; this is cheap insurance against future cap bumps.
- `build_test_single_first_feature_split` narrowed `pub` → `pub(crate)`. It's a test-only fixture; only in-crate callers use it.

Verified with `make test` on 58b — 28+11 tests pass, no regressions. Dependent branches 58c/58d rebased and force-pushed.

johnjezl:
slmos-review pass 3 — addressed in f9f69823:

- Header field offsets named (`OFF_VERSION`, `OFF_SINGLE_TREE_COUNT`, `OFF_CASCADE_N_CLASSIFIERS`, `OFF_CLF_*`). The parser now self-documents which byte offsets correspond to which fields, and a comment block pins the exporter source-of-truth (`slm-os-scheduler-ai/scripts/export_models.py`) to keep them in sync.
- `eval_tree` doc block now spells out the unchecked-indexing invariant: parser-built models are safe by construction (`validate_node` rejects out-of-range children, `parse_roots_*` rejects out-of-range roots); test-only `from_parts_for_test` callers must keep that invariant.
- `from_parts_for_test` narrowed `pub(crate)` → private. Used only inside this module's own `#[cfg(test)] mod tests`; no out-of-module callers.

Skipped: the `parse_roots_u16`/`parse_roots_u32` (and `_nodes_*`) deduplication suggestion. Per the pass-3 note, the current explicit form is more readable than a parameterised helper, so the refactor isn't a clear net win.

Verified with `make test` on 58b — 2453 PASS lines, 0 FAIL. 58c and 58d rebased onto the new tip.

Issue #920, opened 2026-05-16
ai_xgb: 977/1000 divergence between on-device walker and Python TripleClassifier.predict


johnjezl:
## Background

PR (xgb-sched-904-equiv-bench) lands the `bench xgb-equiv` shell verb that replays the sibling-repo verification corpus through the on-device cascade. **First hardware run (pi-5-2, 2026-05-15) found a real divergence:**

| Metric | Value |
|--------|-------|
| Match rate | **23 / 1000** |
| Per-decision latency | 240,747 ns (~240 µs) |
| First mismatch (i=0) | got `(core=1, priority=2, preempt=1)` vs expected `(core=1, priority=1, preempt=0)` |

The walker is `no_std` Rust with no platform-specific code (no NEON, no FP-tier divergence) so a Jetson run would produce identical numbers. This is a software-only divergence between the on-device cascade walker and Python's `TripleClassifier.predict`.

## Why this matters

The cascade walker re-implements XGBoost inference (tree traversal, sigmoid, multiclass argmax) in `runtime/src/ml/xgb_tree.rs` and the cascade plumbing in `runtime/src/sched/xgb.rs`. If predictions diverge from the trainer's output, the `ai_xgb` policy is making decisions based on a different model than the one we trained — the trained DCR numbers (~95.3% mean across 8 scenarios per `eval_summary.csv`) are not what we'd actually get on hardware.

The synthetic 3-classifier cascade in `rust_run_tests` proves the FFI plumbing works (stage → activate → predict → rollback → clear). It does not prove numerical equivalence with a *trained* cascade — which is exactly what this issue closes the gap on.

## Investigation candidates

The shape of the i=0 mismatch (`core` matches, `priority` and `preempt` differ) narrows the search:

1. **Derived feature drift.** `compute_derived` in `runtime/src/sched/xgb.rs:421` mirrors Python's `add_derived_features` term-for-term. A drift in any of the 5 derived features (max_deadline_urgency, core_util_std, deadline_task_count, gpu_should_use, best_cache_fit_core) would feed wrong inputs to **every** classifier. But `core` matched at i=0, suggesting at least the `core_clf` got the right inputs. The priority/preempt classifiers consume `core_clf`'s output appended to the features — a derived-feature bug would surface in `core` first.

[...]

PR #886, opened 2026-05-16
sched/docs: bench sched-policy ai_xgb row + capstone docs (#851)

johnjezl:
## Summary

Final sub-ticket of [#58](https://github.com/SLM-OS/SLM-Operating-System/issues/58)
(XGBoost scheduler policy umbrella). Wires the new `ai_xgb` policy
into `bench sched-policy` and lands the capstone documentation
updates.

- **`bench sched-policy ai_xgb`** — direct call into
  `rust_sched_xgb_predict` (the policy doesn't go through the
  inference_device registry). When no cascade is active the bench
  prints a hint pointing the operator at
  `slm.sched_model_stage("xgboost", path)`.
- **`docs/benchmarks.md`** reframed around four pluggable
  policies. New comparison table includes the XGBoost row with
  the decision-latency cell explicitly marked "captured by
  `bench sched-policy`" so the reader knows to read the latest
  run. The XGB-vs-MLP DCR delta (97% vs 99% from the sibling
  simulator) is cited as informative — *not* a selection criterion
  — consistent with [#848](https://github.com/SLM-OS/SLM-Operating-System/issues/848).
- **`docs/scheduler.md`** extends the architecture-diagram policy
  strip to include `ai_xgb` + `ai_hailo`, plus a new §"`ai_xgb`
  cascade policy" that documents the staging workflow and CPU-id
  clamping behaviour.
- **`docs/lua.md`** documents the previously-undocumented
  `slm.sched_model_*` Lua bindings; `xgboost` is now an accepted
  kind.
- **`docs/fact-sheets/ai-scheduler.md`** updated to show three
  trained policies (MLP, PPO, XGBoost) plus `ai_hailo`. Notes the
  runtime-only XGBoost storage shape.

### What this PR deliberately does NOT do

- **No hardware run on pi-5-2 or jetson-nano-1.** Both boards are
  shared with #56 (SIMD); per the project rule, hardware access
  needs John's confirmation before flashing. The bench harness is
  wired and ready — the on-target row in `docs/benchmarks.md` will
  be populated with real numbers once the boards are confirmed
  free.

- **No bit-equality test loader against
  `expected_actions_xgb.bin`.** The SD-card-fixture replay
  (~half-day shell-verb work) is captured in the fact-sheet's
  "Skipped / Blocked" section. Today's in-kernel coverage is the
  synthetic 3-classifier test in `rust_run_tests` from #885.

### Dependencies

[...]

PR #921, opened 2026-05-16
sched/xgb: bench xgb-equiv verb + Pi 5 hardware run (closes #904)

### LINKED ISSUE #904: ai_xgb: on-target bit-equality test against Python TripleClassifier.predict
## Background

PR #886 (58d) ships the `ai_xgb` cascade scheduler policy and notes in
`docs/fact-sheets/ai-scheduler.md` that an on-target bit-equality test
against the Python trainer's `TripleClassifier.predict` does not yet
exist.

The kernel-side `rust_run_tests` (`runtime/src/lib.rs`) covers a
synthetic 3-classifier cascade today — that exercises the FFI surface
(stage / activate / predict / rollback / clear) but proves nothing
about numerical equivalence between the trained model and the on-device
predictor.

## What's needed

1. Sibling repo `slm-os-scheduler-ai` already produces an
   `expected_actions_xgb.bin` test corpus alongside the trained
   cascade — 1000 ground-truth `(state[108], core, priority, preempt)`
   tuples emitted by Python's `TripleClassifier.predict` over a
   representative input distribution.
2. Add a shell verb (e.g. `bench xgb-equiv <state-corpus> <expected-corpus>`)
   that:
   - reads the corpus from VFS
   - calls `rust_sched_xgb_predict` on each state vector
   - diffs the (core, priority, preempt) triple against the expected one
   - reports first-mismatch index + counts
3. Run on pi-5-2 and jetson-nano-1 with the trained `xgb_sched.smb` staged.

## Acceptance criteria

- All 1000 corpus entries match bit-exactly on at least one platform.
- Any mismatch is investigated rather than waved off — the on-device
  XGBoost engine and the Python predictor must agree, or the cascade
  isn't actually deployable as-is.

## Why P3-low

The synthetic cascade test in `rust_run_tests` already proves the
plumbing. Without this test, a *trained* cascade could silently
deviate from Python predictions due to f32 vs f64 rounding,
threshold-comparison direction, or label-table off-by-one. Worth
having before the policy is recommended for any production-style
demo, but not blocking the merge of #886.

## Related

- #855 — `ai_xgb` policy implementation
- #851 — bench harness + capstone framing
- #850 — sibling-repo binary blob exporter
- #58 — XGBoost scheduler umbrella

johnjezl:
## Summary

Closes [#904](https://github.com/SLM-OS/SLM-Operating-System/issues/904) — the on-target bit-equality verification path against Python's `TripleClassifier.predict`. The verb works as designed: it caught a real divergence on first run.

### What landed
- **`bench xgb-equiv <vectors> <actions>` shell verb** in `kernel/src/shell_sys.c`. Replays the sibling-repo corpus through the staged cascade, reports match count + per-decision latency + first-mismatch detail. Accepts both VFS and FAT (`0:/slmstore/...`) paths.
- **`runtime_blob_read_fat_buf` + `runtime_blob_is_fat_buf_path`** — public wrappers around the file-static `runtime_blob_read_fat_path` so shell verbs can read FAT files without duplicating the f_mount/f_open dance. The Pi 5 LittleFS mount is too small for the 9 MB cascade + 432 KB corpus.
- **QEMU smoke test** `sched_xgb_equiv_loop_synthetic` in `runtime/src/lib.rs::rust_run_tests` — 16-iteration predict loop with varying state, mirrors the bench verb shape. Covers the iteration loop in QEMU even when no trained corpus is staged.
- **Latent bug fix in `bench_xgb_one`**: the `state[i] = 0.0f` literal didn't actually compile under `-mgeneral-regs-only` (which `shell_sys.c` is built with). PR #886 introduced it but the kernel-test build inherited a stale `.obj` from before that PR and masked the failure. Forcing a rebuild here surfaced it; replaced with a `memset` on a `uint8_t[]` buffer (same all-zero IEEE 754 bit pattern, no float literal).
- **`docs/benchmarks.md`**: documented the Pi 5 hardware result and the divergence finding.

### Pi 5 hardware run (2026-05-15, BCM2712 Cortex-A76 @ 2.4 GHz)

| Metric | Value |
|--------|-------|
| Match rate | **23 / 1000** |
| Per-decision latency | **240,747 ns (~240 µs)** |
| First mismatch (i=0) | got `(core=1, priority=2, preempt=1)` vs expected `(core=1, priority=1, preempt=0)` |

The 977/1000 divergence is **real and intentional to surface**. The cascade walker is pure software (`no_std` Rust, no NEON or platform-specific FP), so a Jetson run would produce identical numbers — skipped for that reason. **Investigation tracked in [#920](https://github.com/SLM-OS/SLM-Operating-System/issues/920) (P1-high)** with a candidate-list narrowing the search.

### Why land the verb without the fix

The verb is what #904 explicitly requested. With it landed:
- The divergence is now visible and reproducible — you can `bench xgb-equiv` after any future cascade-walker change and instantly see whether the change moved the needle.
- The next person investigating #920 has a one-command regression check.
- The `ai_xgb` policy doesn't actually become *worse* by landing this — it has the same divergence today, just unmeasured.

## Test plan
- [x] `make test AI_SCHED=ON` clean — 7/7 `sched_xgb_*` tests pass (1 new), same 8 pre-existing failures as main.
- [x] Pi 5 hardware bench run captured (above).
- [ ] Jetson hardware bench — skipped per the design rationale (pure-software walker).
- [ ] 1000/1000 match — gated on #920 fix (separate PR).

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -297,6 +297,56 @@ clamps via modulo and continues, with a one-shot WARN. The
 distribution may skew on 4-core platforms but no decision is dropped
 to the heuristic fallback.
 
+#### Bit-equality verification (`bench xgb-equiv`)
+
+The sibling repo's exporter emits two corpus files alongside
+`xgb_sched.smb`: `test_vectors_xgb.bin` (1000 × 108 f32 input states)
+and `expected_actions_xgb.bin` (1000 × 3 i32 ground-truth `(core,
+priority, preempt)` triples produced by Python's `TripleClassifier
+.predict`). Replay the corpus through the on-device cascade with:
+
+```
+# In the SLM-OS shell, after the stage/activate step above:
+bench xgb-equiv 0:/slmstore/test_vectors_xgb.bin 0:/slmstore/expected_actions_xgb.bin
+```
+
+The verb walks both files in lockstep, runs `rust_sched_xgb_predict`
+on each state, diffs the triple against the expected one, and prints
+a pass/fail summary plus the first-mismatch index + actual-vs-expected
+triple if any disagreement appears. Exit code is 0 on full match,
+1 on any mismatch — useful for scripted hardware verification. Both
+VFS-rooted (`/mnt/files/...`) and FAT-rooted (`0:/slmstore/...`)
+paths are accepted.
+
+This exists because the on-device walker re-implements XGBoost
+inference in `no_std` Rust (tree traversal, sigmoid, multiclass
+argmax). The synthetic 3-classifier cascade in `rust_run_tests`
+proves the FFI plumbing works; this verb proves numerical equivalence
+with the trainer on a *trained* cascade. Closes
+[#904](https://github.com/SLM-OS/SLM-Operating-System/issues/904).
+
+**Pi 5 result (2026-05-15, BCM2712 Cortex-A76 @ 2.4 GHz):**
+
+| Metric | Value |
+|--------|-------|
+| Match rate | **23 / 1000** |
+| Per-decision latency | **240,747 ns (~240 µs)** |
+| First mismatch (i=0) | got `(core=1, priority=2, preempt=1)` vs expected `(core=1, priority=1, preempt=0)` |
+
+The 977/1000 mismatch rate is a real divergence between the on-device
+walker and Python's `TripleClassifier.predict` — exactly the class of
+bug this verb was designed to catch. The shape (`core` matches at
+i=0, `priority` and `preempt` differ) is consistent with a
+derived-feature drift, label-encoder inverse-transform mismatch, or
+classifier-input-vector ordering bug. **Investigation tracked in
+[#920](https://github.com/SLM-OS/SLM-Operating-System/issues/920)**;
+the verb itself is sound and will be the canonical regression-detection
+tool once the divergence is fixed.
+
+The cascade walker is pure software (no NEON, no FP-tier divergence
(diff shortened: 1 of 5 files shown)


johnjezl:
slmos-review polish — addressed in 268ecd40:

- `runtime_blob_file`: renamed `runtime_blob_is_fat_buf_path` → public `runtime_blob_is_fat_path`. The "buf" suffix existed only to avoid colliding with a now-deleted file-static; one public symbol is cleaner.
- `shell_sys.c::bench_xgb_one`: comment now flags the `aligned(4)` assumption (sufficient for scalar f32 loads, bump to `aligned(16)` if NEON quad-word loads are added later).
- `shell_sys.c::bench_xgb_equiv`: extracted `xgb_equiv_read_corpus` helper. Dedups the two near-identical FAT-vs-VFS read blocks. Function dropped from ~155 lines to ~100.
- `shell_sys.c::bench_xgb_equiv`: cleanup path uses a single `goto cleanup` label with a final `pmm_free_pages` pair. Removes the 4 duplicated cleanup paths flagged in the review.
- `runtime/src/lib.rs::sched_xgb_equiv_loop_synthetic`: stage / activate return codes now checked. Failure has locality (`equiv_ok = false` from `stage_rc != 0`) instead of cascading into a misleading "loop mismatched" diagnostic.

QEMU `make test AI_SCHED=ON` clean — 7/7 `sched_xgb_*` tests pass, same 8 pre-existing baseline failures.

Issue #216, opened 2026-04-16
Pi 5: secondary CPUs intermittently dormant after boot (work-stealing test flake root cause)


johnjezl:
## Summary

On Pi 5 (`pi-5-1`), one or more of CPU 1 / CPU 2 / CPU 3 is intermittently "dormant" during the integration test run — `sched_diag_schedule[cpu]` never increments for the entire test window, meaning the CPU never enters `schedule()` even though tasks are being queued onto it with `scheduler_add_task_to_cpu()` (which issues SEV via `smp_notify_cpu`). Which CPUs are affected varies boot-to-boot.

Observed during work-stealing test flakiness investigation (branch `worktree-test-flakiness-fix`). Data across 5 boots with per-attempt diagnostics:

| Boot | CPU 1 | CPU 2 | CPU 3 | Notes |
|---|---|---|---|---|
| 1 | dead | dead | dead | 8 multi-core tests FAIL |
| 2 | dead | alive | dead | Only CPU 2 picks up work |
| 3 | dead | dead | alive | Only CPU 3 picks up work |
| 4 | dead | alive | dead | `test_task_migration` FAIL (CPU 3 dead) |
| 5 | dead | dead | alive | Only CPU 3 picks up work |

"dead" = `sched_diag_schedule[c]` delta over the 9 ms test attempt window == 0, and `sched_diag_steal_attempts[c]` == 0.

## Evidence

Example of a \"dead CPU 1\" attempt:

\`\`\`
ws-diag attempt=0 distinct=1 elapsed_us=9324 cpus=[2,2,2,2,2]
  per-cpu(att/succ/stale/sched/picked)=
  0/0/0/3376/6752,   # CPU 0 (test driver, yield loop)
  0/0/0/0/0,         # CPU 1 — never schedules, never idles
  2676/5/0/2676/6,   # CPU 2 — healthy stealer
  0/0/0/0/0          # CPU 3 — never schedules, never idles
\`\`\`

CPU 1 at boot passes \`test_smp_notify_cpu_wakes_secondary\` (runs a probe task on CPU 1). So CPU 1 is initially alive — it goes dormant between \`test_smp_notify_cpu_wakes_secondary\` completing and \`test_work_stealing_distributes_load\` starting.

## Impact

[...]

PR #948, opened 2026-05-16
msg_router: reject-when-busy + flow-control counters (#869)

### LINKED ISSUE #869: msg_router: single-slot mailbox can silently drop messages under back-to-back delivery
## Description

Surfaced by the #67 concurrency-test audit. The `Mailbox` struct in `runtime/src/msg_router.rs:60-66` is a single-slot structure (one set of `ready/ack/data/topic/priority` atomics). When two `publish_internal` calls target the same mailbox before the subscriber has acked the first message, the second `deliver()` overwrites the first message's data, topic, and priority in place. The subscriber then sees the second message; the first is silently lost.

This is distinct from the LAST_RECEIVED targeting case covered by #867 (where two messages arrive in *different* mailboxes for the same component — LAST_RECEIVED ensures ack targets the right one). The drop here happens *within a single mailbox*.

## Reproducer

Conceptual sequence (#864's prio-ordering test will likely surface this under sustained load):

1. Subscribe component C to topic `/x`.
2. Publisher emits message M1 to `/x` (priority 0). Mailbox state: `data=M1, ready=1, ack=0, priority=0`.
3. Publisher emits message M2 to `/x` (priority 5) before C calls `receive()`. `deliver()` runs: writes `data=M2, topic=/x, ack=0, priority=5, ready=1` (the existing `ready=1` is overwritten with `ready=1`). M1 is gone.
4. C calls `receive()` — returns M2.
5. C acks. Mailbox cleared.

C never saw M1.

Note that `publish_internal` (`runtime/src/msg_router.rs:610`) waits for ack before returning, so a single-publisher serial workload can't hit this. The drop requires **multiple publishers running concurrently** or a publisher running on a CPU while the subscriber's CPU is busy and hasn't reached the ack yet.

## Impact

- Existing test coverage doesn't surface the drop because tests have one publisher waiting on ack before publishing the next message.
- Production paths where multiple components publish to the same topic from different CPUs (e.g. telemetry feeds, sensor fan-in) can lose data silently.
- The router's return value (`delivered`) increments only when an ack is observed — so a dropped message returns 0 to its publisher. This is at least *observable*, but the dropped message has already been overwritten in the mailbox by the time the publisher sees `delivered=0`.

## Proposed fixes (out of scope for this ticket; for the issue body)

1. **Queued mailboxes** — bounded ring buffer per mailbox (e.g. 4-8 slots). Each `deliver()` appends; `receive()` pops. Adds a per-mailbox lock / sequence counter. Higher memory overhead per subscription.
2. **Reject second deliver while first is unacked** — `publish_internal` checks `ready` before calling `deliver`; if set, either spins on ack or returns a "would block" code to the publisher. Preserves zero-loss semantics; cost is publisher latency.
3. **Caller-side serialization** — document that callers must serialize publishes to the same topic. Cheapest but pushes burden up. Probably not acceptable for a kernel pub/sub primitive.

Recommend (2) for simple cases (one ack-wait per delivery already exists in `publish_internal`), with (1) as a follow-on if measured publisher latency under contention is too high.

## Scope

1. Decide on the fix (queued vs reject-when-busy vs caller-serialize). Document the decision in this issue.
2. Implement.
3. Add a regression test (the conceptual reproducer above, with concurrent publishers on separate CPUs).
4. Update documentation (`docs/architecture.md` IPC section, `runtime/CLAUDE.md`).

## Not on Capstone milestone

This is a real bug but the fix is non-trivial and the workaround (one publisher per topic, ack-then-publish discipline) is achievable in current code paths. Filed for post-capstone follow-up. Tests in #864 / #867 will document the limitation; this ticket closes it.

## References

- Surfaced by the #67 audit (2026-05-15).
- Mailbox struct: `runtime/src/msg_router.rs:60-66`.
- `Mailbox::deliver`: `runtime/src/msg_router.rs:104-110`.
- `publish_internal`: `runtime/src/msg_router.rs:610-692`.
- Related: #66 (preemption-safe spinlock — closed), #67 (concurrency stress tests — in progress).

johnjezl:
## Summary

Closes #869 via option 2 (reject-when-busy + publisher retry) from the [scope comment](https://github.com/SLM-OS/SLM-Operating-System/issues/869#issuecomment-4468136553).

Pre-fix: two `publish_internal` calls targeting the same mailbox before the subscriber acked the first silently overwrote the first message — both publishers observed the ack and reported `delivered=1`. Lost data, no observable error.

Post-fix: `Mailbox::try_deliver` refuses to overwrite an unacked mailbox. The ack-waiting path (`msg_router_publish` / `_priority`) yields and retries inside the existing `ACK_TIMEOUT_SECS` budget — zero-loss preserved. The nowait path (`msg_router_publish_nowait`) drops and counts — the caller sees the failed delivery in the return value instead of silent loss.

## Changes

**Runtime (`runtime/src/msg_router.rs`)**:
- `Mailbox::deliver` → `Mailbox::try_deliver` returning `bool`.
- `publish_internal`: yield-and-retry on busy mailbox within `ACK_TIMEOUT_SECS`; bumps `FLOW_CONTROL_BUSY_RETRIES`.
- `publish_internal_nowait`: drops on busy mailbox; bumps `FLOW_CONTROL_NOWAIT_DROPS`; failed target excluded from `delivered` return.
- New FFI: `msg_router_flow_control_stats(busy_out, drops_out)` + `msg_router_flow_control_reset()`.

**Tests**:
- `test_msg_router.c::test_publish_nowait_basic_contract` — existing canary, updated. Previously asserted "second nowait publish overwrites the first as #869 documents the limitation"; now asserts "second nowait publish is rejected, `NOWAIT_DROPS` increments, subscriber receives the first message." This is the test the original comment explicitly flagged for update when #869 lands.
- `test_integration.c::test_msg_router_same_mailbox_zero_loss` — NEW. Two publishers on CPUs 1 and 2 emit to the same topic; subscriber on CPU 3 drains. Invariant: subscriber-seen counts == publishers' `delivered` counts.

**Documentation**:
- `runtime/CLAUDE.md` — replaced the "Single-slot mailbox limitation" stub with the new "Mailbox flow control (#869, reject-when-busy)" section. Added "Concurrent-publisher TOCTOU caveat" noting the residual smaller race window (two CPUs both see `ready==0` simultaneously) that only a CAS-loop or queued-mailbox redesign would close.
- `docs/architecture.md` — replaced the "Known limitation" bullet with the new flow-control contract description.

## Test plan

- [x] QEMU virt: `make test` green
- [x] `make kernel PLATFORM=RASPI5`: clean
- [x] `make kernel PLATFORM=JETSON_ORIN_NANO`: clean
- [ ] Hardware revalidation on pi-5-2 or jetson-nano-1 — QEMU's serial TCG hides the publisher race (publishers don't actually run concurrently under TCG), so the new `test_msg_router_same_mailbox_zero_loss` is QEMU-passing-by-accident. Real SMP verification needs hardware. The canary in `test_msg_router.c` does fire on QEMU.

## Notes / out of scope

- A residual TOCTOU window exists where two CPUs racing into `try_deliver` for the same mailbox both observe `ready==0` before either's `ready.store(1)`. The `try_deliver` check closes the larger "deliver-then-deliver-without-yield" window that single-publisher sequential code hits; the smaller cycle-aligned window remains. Closing it requires a CAS-loop redesign (treat `ready` as 0=idle, 2=writing, 1=ready) or the full queued-mailbox option from #869's scope. Both are strictly larger changes; not needed for in-tree workloads today. If a telemetry fan-in or sensor producer trips it on hardware, escalate to one of those.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -191,12 +191,24 @@ Pub/Sub" for the full reference):
   component, even when a newer message lands in a *different* mailbox
   between the receive and the ack. The bookkeeping survives genuine
   cross-CPU activity (publish on one CPU while ack runs on another).
-- **Known limitation — single-slot mailbox overwrite.** A *single*
-  mailbox can lose a message if two `publish_internal` calls target it
-  back-to-back with no intervening ack. The standard ack-waiting publish
-  loop blocks between iterations so a single publisher cannot trigger
-  this; the case requires multiple concurrent publishers writing to the
-  same subscriber's mailbox. Tracked for post-capstone follow-up as #869.
+- **Mailbox flow control (#869, reject-when-busy).** `Mailbox::try_deliver`
+  refuses to write while a prior message is still unacked. The
+  ack-waiting `msg_router_publish` / `msg_router_publish_priority` path
+  yields and retries inside the same `ACK_TIMEOUT_SECS` budget, so the
+  zero-loss contract holds for concurrent publishers writing to the same
+  mailbox. The fire-and-forget `msg_router_publish_nowait` path drops
+  the message instead of overwriting and excludes the target from its
+  `delivered` return value, so the caller can see that the send didn't
+  land. Two counters — `flow_control_busy_retries` (ack-waiting path
+  yielded and retried; no loss) and `flow_control_nowait_drops`
+  (nowait path dropped a message; caller opted out of blocking) — are
+  exposed via `msg_router_flow_control_stats()` and the shell
+  `msg stats` command. A residual TOCTOU window remains where two
+  CPUs racing into `try_deliver` for the same mailbox can both see
+  `ready == 0` between the Acquire load and the Release stores;
+  closing it requires a CAS-loop redesign or the full queued-mailbox
+  option in #869's scope, neither of which is needed for the workloads
+  in tree today.
 
 ### SMP Support
 
--- a/kernel/tests/test_integration.c
+++ b/kernel/tests/test_integration.c
@@ -2001,6 +2001,17 @@ static void test_msg_router_ack_targets_last_received_cross_cpu(void)
  *
  * Requires cpu_count >= 4 (publisher A on CPU 1, publisher B on CPU 2,
  * subscriber on CPU 3, plus shell on CPU 0).
+ *
+ * IMPORTANT — QEMU vs hardware: this test passes on QEMU virt regardless
+ * of whether the fix is present, because QEMU's serial TCG scheduler
+ * interleaves the publishers and subscriber deterministically and never
+ * opens the race window in the first place. The test's primary value is
+ * on real Pi 5 / Jetson SMP hardware where publishers genuinely run
+ * concurrently. A green QEMU run is structural only — do NOT conclude
+ * from it that the underlying mailbox semantics are correct without a
+ * hardware run. The matching same-CPU contract canary in
+ * test_msg_router.c (`test_publish_nowait_basic_contract`) DOES fire on
+ * QEMU and is the day-to-day regression guard.
  * ============================================================================ */
 
 #define SM_C_COMPONENT 24
--- a/kernel/tests/test_integration.c
+++ b/kernel/tests/test_integration.c
@@ -120,6 +120,14 @@ static struct {
 /* msg_router #67c: cross-CPU LAST_RECEIVED ack-targeting test. */
 #define NC_MSGAR_TRIGGER     31  /* main → helper: publish /b now */
 #define NC_MSGAR_HELPER_DONE 32  /* helper → main: publish complete */
+/* msg_router #869: same-mailbox concurrent-publisher zero-loss test. */
+#define NC_MSGSM_A_DELIVERED 33  /* pub_a's total delivered count */
+#define NC_MSGSM_B_DELIVERED 34  /* pub_b's total delivered count */
+#define NC_MSGSM_SUB_A_SEEN  35  /* subscriber's count of A-payloads */
+#define NC_MSGSM_SUB_B_SEEN  36  /* subscriber's count of B-payloads */
+#define NC_MSGSM_PUB_A_DONE  37
+#define NC_MSGSM_PUB_B_DONE  38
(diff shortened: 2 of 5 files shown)


johnjezl:
## Hardware revalidation on pi-5-2 — PASS

Deployed via \`sdwire_update\` (copy \`build/kernel-test/slmos.bin\` → \`kernel_2712.img\` + reboot). Pi 5 boots into test kernel, runs all suites, reaches shell.

### test_msg_router_same_mailbox_zero_loss — PASS

\`\`\`
[TEST] kernel/tests/test_integration.c
  [PASS] test_multicore_basic
  [PASS] test_task_migration
  [PASS] test_stress_multicpu
  [PASS] test_lock_contention
  [PASS] test_task_lifecycle
  [PASS] test_nc_memory_accessible
  [PASS] test_nc_alloc_alignment
  ...
  [PASS] test_msg_router_cross_cpu
  [PASS] test_msg_router_multi_subscriber
  [PASS] test_msg_router_priority_concurrent
  [PASS] test_msg_router_ack_targets_last_received_cross_cpu
  [PASS] test_msg_router_same_mailbox_zero_loss      ← the new #869 test
  [PASS] test_smp_n…
\`\`\`

All msg_router tests PASS under real Pi 5 SMP — two publishers on CPUs 1 and 2 publishing 5 messages each to \`/sm\`, subscriber on CPU 3 received all 10 (both A_seen=5 and B_seen=5, both pub_delivered=5). The reject-when-busy contract holds under genuine concurrency, not just QEMU's serial TCG.

### Notes

- Pi 5 #216 dormancy interjected on CPUs 1, 2, and 3 during early tests (\`cpu_supervisor\` resurrection ran \`ALREADY_ON\` workaround). Recovery succeeded; no regression caused by this PR. Same intermittent pattern observed on every recent Pi 5 hardware run.
- The contract canary in \`test_msg_router.c::test_publish_nowait_basic_contract\` was already green on QEMU and is the day-to-day regression guard.
- pi-5-2 board released.

Ready to merge.

Issue #961, opened 2026-05-17
AI Eviction: tree-prune XGBoost + batch MLP to hit <1 µs hot-path target


johnjezl:
## Goal

Bring `xgboost`, `mlp`, and `cacheus` eviction policies under the
**< 1 µs per `select_victim` call** target from the Phase AI-Eviction
TODO (`docs/archive/todos/TODO - PHASE AI-Eviction.md`) on real Cortex-A78
silicon.

## Background — what we measured

Hardware bench captured 2026-05-17 via `test_bench_all_policies` on
`pi-5-2` (BCM2712 Cortex-A76 @ 2.4 GHz) and `jetson-nano-1`
(Cortex-A78AE @ 1.5 GHz, NS-EL2/VHE) — see `docs/eviction.md` §"Latency
Benchmarks (M9)" for the full table.

Classical policies (`first_candidate`, `lru`, `lfu`, `arc`, `slm`) all
stay under 1 µs across both platforms and both build configs. The ML
policies miss:

| Policy | Pi 5 default | Pi 5 MODELS | Jetson default | Jetson MODELS |
|--------|---:|---:|---:|---:|
| xgboost | 4,117 ns | 15,029 ns | 6,098 ns | 21,956 ns |
| mlp | 4,157 ns | 129,151 ns | 6,170 ns | 174,248 ns |
| cacheus (ml_only) | 8,999 ns | 145,061 ns | 13,311 ns | 197,228 ns |

The `EVICTION_MODELS=ON` column is the only column that runs the
trained weights; `default` runs the stub generators and so under-
samples the real work.

## Levers (from sibling-project Phase 5)

1. **Tree-pruning the XGBoost ensemble** — the imported model is 200
   trees deep; the sibling project's Phase 5 work found a much smaller
   ensemble (~16 trees) produced ≥99 % equivalent decisions on the
   eviction corpus, cutting `xgb_predict` latency by ~12×.
2. **Batch-inference optimisation for the MLP** — currently every
   `select_victim` call runs the int8 forward pass once per candidate;
   batching 8 candidates into one matmul/quant pass amortises the
   per-call dispatch overhead. Sibling project measured ~5× on
   Cortex-A78 with `vqdmlal_s16` paths.

Both levers feed into CACHEUS via the underlying experts.

## Done when

- xgboost < 1 µs / call on pi-5-2 with `EVICTION_MODELS=ON` (Pi 5 is
  the easier target because of clock + cache size).
[...]

Issue #957, opened 2026-05-17
EVICTION_MODELS=ON build fails: unused_imports in generated xgb_policy_generated.rs


johnjezl:
## Repro

Default branch (`main` @ `37fe2bae` as of this filing), clean tree:

```
make test EVICTION_MODELS=ON
```

Fails with:

```
error: unused import: \`crate::mm::eviction::policy::BlockFeatures\`
 --> src/mm/eviction/generated/xgb_policy_generated.rs:5:5
  |
5 | use crate::mm::eviction::policy::BlockFeatures;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: \`-D unused-imports\` implied by \`-D warnings\`

error: could not compile \`slm-runtime\` (lib) due to 1 previous error
```

## Root cause

The auto-generated file imports `BlockFeatures` but only uses the array type alias `&[f32; 27]`:

```rust
use crate::mm::eviction::policy::BlockFeatures;

pub fn xgb_predict(features: &[f32; 27]) -> f32 {
```

The `BlockFeatures` type alias is `[f32; 27]` (`runtime/src/mm/eviction/policy.rs:45`) — same type, but the import is technically unused, and the runtime crate compiles under `-D warnings`.

Generated on 2026-04-13 by `scripts/import_eviction_weights.sh` per the file header. Default build (no `EVICTION_MODELS=ON`) substitutes `xgb_stub.rs` which doesn't have the import — so the break only surfaces on `EVICTION_MODELS=ON` builds.

## Fix

The generator template (`slm-os-page-sim/scripts/export_to_slmos.py` or whatever emits this file) should either:

- Drop the `BlockFeatures` import line — the signature already uses the array form directly, and downstream consumers don't need it from this module.
- **Or** swap the signature to `pub fn xgb_predict(features: &BlockFeatures) -> f32` so the import is actually used.

Either fix should ship from the sibling repo so the generated artifact is correct at source. As an interim stopgap, an `#[allow(unused_imports)]` at the top of `xgb_policy_generated.rs` works but will be erased on next regeneration.

## Scope

- Does NOT block #932 (consumer-side bundle uses `runtime_xgboost::parse_payload` + the FFI, not `generated::xgb_predict`). Default `make test` still passes.
- Does affect anyone running `make test EVICTION_MODELS=ON` or shipping a build with the trained weights baked in.
- Affects all platforms.

## Related

[...]

Issue #953, opened 2026-05-17
eviction: evaluate promoting xgboost from opt-in to default policy


johnjezl:
## Motivation

XGBoost eviction is currently opt-in (`SLM_DEFAULT_EVICTION_POLICY=xgboost`); the boot-time default is LRU (\`runtime/src/mm/eviction/registry.rs:135\`). XGBoost benchmark numbers suggest it is a strong candidate for THE default, but the move has not been formally evaluated.

This is a **decision-gate ticket**, not an implementation ticket. The code change itself is a one-line flip in `default_policy()`. What's needed is the supporting evidence to make that flip with confidence — and the explicit recognition that SLM-OS keeps all policies pluggable regardless (per #848).

## Scope

Produce a short decision memo (`docs/decisions/xgb-as-default-eviction.md` or similar) summarizing:

- **Correctness:** on-device vs. trainer equivalence, from #932's `bench xgb-equiv-evict`.
- **Quality:** hit-rate vs. LRU/ARC/CACHEUS on held-out workloads, from the sibling-repo eval harness (`slm-os-page-sim` companion ticket).
- **Latency:** per-platform decision latency, from #108 (Pi 5), #109 (Jetson), #110 (x86-64).
- **Concurrency:** behavior under multi-CPU stress, from #116.
- **Footprint:** blob size in workspace pool, eviction-path code size.
- **Risk:** failure mode when blob is missing or rejected (fallback today is LRU — confirm it still is, and test it).

If the memo approves the flip:
- One-line change to `default_policy()`.
- Update the `default_is_lru()` test at `runtime/src/mm/eviction/registry.rs:421`.
- Doc updates anywhere LRU-as-default is asserted.

## Acceptance

- Decision memo committed with measured numbers (not hand-wave \"xgb is fast\").
- Decision is explicit: flip / don't flip / revisit when X.
- If \"flip\": registry change + test update + doc sweep, all in one PR.
- Pluggability preserved — LRU/ARC/CACHEUS/MLP remain selectable via `SLM_DEFAULT_EVICTION_POLICY` and the runtime `set_eviction_policy` API.

## Dependencies (blocks this ticket)

- #932 — equivalence verb (correctness)
- #108 / #109 / #110 — per-platform latency benchmarks
- #116 — multi-CPU concurrent alloc/eviction stress
- Sibling-repo \"eviction quality eval harness\" ticket (filed alongside this one)

## Out of scope

- Removing other eviction policies — see #848.
[...]

Issue #932, opened 2026-05-16
ai_xgb eviction: add bench xgb-equiv-evict for on-device vs trainer numerical equivalence


johnjezl:
## Motivation

The scheduler cascade has `bench xgb-equiv` (issue #904, PR #921) which replays a verification corpus through the on-device walker and compares against Python's `TripleClassifier.predict`. That verb caught a real divergence — #920, 23/1000 match — that we would otherwise have shipped.

The XGBoost **eviction** policy uses the same `runtime/src/ml/xgb_tree.rs` engine but has no equivalent equivalence check. "XGBoost eviction matches Python" is currently an assumption, not a measurement. Given that XGBoost eviction is a strong candidate for THE default eviction policy in SLM-OS based on its performance numbers, that assumption needs verification.

## Scope

Add `bench xgb-equiv-evict <vectors-path> <expected-path>` (or similar name) mirroring the scheduler verb:

- Routes FAT (`0:/...`) and VFS (`/mnt/files/...`) paths via existing `runtime_blob_read_fat_buf` / VFS helpers.
- Replays N input vectors through `rust_eviction_blob_predict` (or whatever the activation-side entrypoint becomes).
- Compares per-vector output against an `expected_*.bin` corpus produced by the trainer.
- Reports match-rate, first-mismatch index, ns/predict.
- Returns 0 on PASS, 1 on FAIL.

Corpus generation on the trainer side: extend `scripts/export_models.py` (sibling repo) to emit `test_vectors_xgb_evict.bin` + `expected_evict.bin` alongside the model blob.

## Acceptance

- `bench xgb-equiv-evict 0:/slmstore/test_vectors_xgb_evict.bin 0:/slmstore/expected_evict.bin` reports `N/N match PASS` on pi-5-2 (and ideally also on the synthetic-cascade test in `runtime/src/lib.rs::rust_run_tests`).
- Documentation entry in `docs/benchmarks.md` matching the existing scheduler section.
- A small embedded corpus (≤16 vectors) added to `rust_run_tests` so future regressions fail in `make test` rather than on hardware.

## Relationship to #920

If #920's root cause turns out to be in the shared `xgb_tree` engine (candidate 3: f32/f64 threshold drift), this verb is **required to verify the eviction fix**. If #920's root cause is cascade-specific (candidates 1, 2, 4 — leading suspicion), this verb is **forward-looking insurance** confirming the eviction path is clean and catching future regressions.

[...]

PR #967, opened 2026-05-17
eviction: XGBoost graduation pathway — docs + schema guard + --baked verb (closes #952)

### LINKED ISSUE #952: ai_xgb eviction: graduation pathway needs docs + safety nets
## Background

SLM-OS's eviction `XGBoostPolicy` supports two model sources: a baked function (`runtime/src/mm/eviction/generated/xgb_policy_generated.rs`) compiled into the kernel, and a dynamically-loaded SEMB blob via `rust_eviction_blob_stage` + `RuntimeXGBoostModel::predict`. The intended lifecycle is "experiment via the blob path, then graduate the winner into the baked path for the next official kernel image."

The runtime supports the consumer side of this lifecycle cleanly: `XGBoostPolicy::score_row` checks for a staged blob first and falls back to the baked function when none is present. What's not yet supported:

1. **No documented graduation procedure.** A developer doing it for the first time has to reverse-engineer the steps from `scripts/import_eviction_weights.sh`'s header comment and intuition about page-sim's exporter. There's no `docs/` entry, no checklist, no warning about the gotchas below.

2. **No compile-time guard on feature ordering.** Both `RuntimeXGBoostModel` (`FEATURE_COUNT = 27`) and the generated `xgb_predict()` (`features: &[f32; 27]`) hardcode the layout. The if/else chain references features by index with name-only comments (`features[13] /* predicted_reuse_dist */`). If page-sim's `FeatureConfig` ever re-orders or inserts a feature, the baked code silently uses the wrong index at every node. Nothing in the build catches this.

3. **No documented compile-flag boundary.** `xgb_policy_generated.rs` is full of `f32` literals (`0.199489996_f32`). Eviction code today isn't built under `-mgeneral-regs-only` (that's shell-only), but the boundary isn't recorded anywhere. If a future change pulls eviction into a no-FP-literals build path, the file silently stops compiling at graduation time, not at flag-flip time.

4. **Test fixtures that hardcode model output break on graduation.** Tests under `kernel/tests/test_eviction*.c` and `runtime/src/mm/eviction/tests/` may assert specific scores against specific inputs. New baked model invalidates those fixtures; developer has to regenerate them. There's no script or instruction for doing so.

5. **No on-device equivalence check after graduation.** Page-sim's `verify_predictions` runs the generated code's logic through a Python interpreter and compares against the trained model — but that's still Python comparing Python. It doesn't verify that the actual ARM64/x86_64 baked code (with whatever FP rounding the kernel compiler applied) matches the source model on real hardware. A regression in compiler tree-walk or FP precision would slip through. A `bench xgb-equiv-evict --baked` sanity verb would catch this.

### Phase 1 — Documentation (covers items 1, 2, 3, 4)

- Add `docs/eviction-xgboost-graduation.md` walking through the full procedure: train candidate → blob-test → run page-sim exporter → regenerate `xgb_policy_generated.rs` → commit → rebuild → flash → re-run regression tests. Include the gotcha list as inline warnings.
- Cross-reference from `docs/scheduler.md` § eviction policies and from `runtime/src/mm/eviction/generated/mod.rs`'s top comment.
- Add a `## Feature schema invariants` section that pins the (FEATURE_COUNT, feature_names, feature_order) tuple as the SLM-OS source-of-truth and warns that page-sim's `FeatureConfig` must match it.

### Phase 2 — Compile-time safety nets (covers item 2, partial)

- Have page-sim's exporter emit a header constant block at the top of `xgb_policy_generated.rs` capturing `(N_FEATURES, feature_names[0..N])`.
- Add a Rust `_assert` or `const _: () = assert!()` block in `runtime/src/mm/eviction/generated/mod.rs` that compares against the SLM-OS-side `BlockFeatures` constants. Mismatch breaks the build with an attributable error rather than silently mis-indexing at runtime.
- (Feature *count* drift is explicitly out of scope per maintainer call — only ordering/naming drift covered here.)

### Phase 3 — Post-graduation on-device equivalence (covers item 5)

- Once #932 lands the blob-path equivalence verb (`bench xgb-equiv-evict`), extend it with a `--baked` mode that calls `generated::xgb_predict()` instead of the blob-loaded predictor. Same corpus comparison, different target. Run as the last step of the graduation procedure.

## Out of scope

- Feature count drift (maintainer explicitly deferred).
- Building the SEMB exporter for the blob path (that's #932's territory).
- Changing the if/else generation format itself.

## Acceptance

- A first-time graduator can follow `docs/eviction-xgboost-graduation.md` end-to-end without reading any source code.
- Re-ordering a feature in page-sim's `FeatureConfig` without updating SLM-OS produces a compile-time error, not a runtime regression.
- Compile-flag boundary for `xgb_policy_generated.rs` is documented in at least one of: the file header, `kernel/CLAUDE.md`, or `docs/eviction-xgboost-graduation.md`.

## Related

- #932 — eviction-side `bench xgb-equiv-evict` (blob path; this issue's Phase 3 builds on it)
- #920 — scheduler equivalence verb that surfaced an analogous on-device-vs-Python divergence
- #58 — XGBoost scheduler umbrella

johnjezl:
## Summary

Closes #952 (all three phases).

### Phase 1 — Documentation
- New \`docs/eviction-xgboost-graduation.md\` (~280 lines) — full lifecycle from blob-path experimentation to baked-in default: when to graduate, 6-step procedure, 5 gotchas, feature schema invariants, related tickets.
- Cross-refs added to \`docs/eviction.md\` §Importing Trained Weights and \`runtime/src/mm/eviction/generated/mod.rs\` module doc.

### Phase 2 — Compile-time feature-schema guard
- Sibling generators (\`slm-os-page-eviction\` PR, \`slm-os-page-sim\` PR) emit \`pub const GENERATED_FEATURE_NAMES: [&str; N]\` alongside \`xgb_predict\`.
- New \`const _: () = { … }\` block in \`generated/mod.rs\` (gated on \`ai_eviction_models\`) compares against runtime \`FEATURE_NAMES\` element-by-element. Mismatch breaks the build with a \`panic!\` in const context.
- **Verified the guard fires.** Temporarily corrupting \`FEATURE_NAMES[0]\` produces:
  \`\`\`
  error[E0080]: evaluation panicked: XGBoost generated feature name differs from runtime FEATURE_NAMES at some index — diff page-sim's FeatureConfig.feature_names against runtime/src/mm/eviction/features.rs::FEATURE_NAMES (see docs/eviction-xgboost-graduation.md §Feature schema invariants)
  \`\`\`
- Reconciled 6 pre-existing label drifts (\`gpu_mapped\` → \`is_gpu_mapped\`, etc.) so the guard passes today. Indices/meanings already aligned; only labels differed. Affects only the \`eviction features\` shell verb output.

### Phase 3 — \`bench xgb-equiv-evict --baked\` mode
- New \`rust_eviction_xgb_predict_baked_compare\` FFI in \`runtime/src/lib.rs\` calls \`generated::xgb_predict\` directly (skipping active-blob lookup), gated on \`ai_eviction_models\`. Returns -5 when off.
- Shell verb extended to accept \`--baked\` as an optional first arg. PASS/FAIL annotated with mode: \`xgb-equiv-evict[baked]:\` vs \`[blob]:\`.
- Purpose: catch ARM64/x86_64 f32 codegen drift the blob-path check can't see (toolchain upgrade, accidental \`-ffast-math\`, LTO interaction). The graduation doc's Step 5 / Gotcha #5 is the rationale.

## Verification

- [x] \`make test AI_SCHED=ON\` — clean, same 8 unrelated pre-existing failures
- [x] \`make test EVICTION_MODELS=ON\` — ALL TESTS PASS (this build variant doesn't include the 8 unrelated failing tests)
- [x] Guard mutation test — panics at the right site with the right diagnostic
- [ ] Pi 5 hardware: \`bench xgb-equiv-evict --baked\` against 1000-vector corpus (in progress this session)

## Companion PRs

The schema guard is two-sided. Sibling-repo PRs add the matching generator constants:
- \`slm-os-page-eviction\` PR (\`emit-feature-names-const\`)
- \`slm-os-page-sim\` PR (\`emit-feature-names-const\`, shares the SLM-OS-org PR namespace with page-eviction via fork link)

Without the sibling PRs, the next \`scripts/import_eviction_weights.sh\` run loses the \`GENERATED_FEATURE_NAMES\` constant and the SLM-OS build breaks. Merge order doesn't matter much; the SLM-OS-side guard is dormant when the constant is missing only because the const_assert would also fail on missing-symbol — so this PR's merge effectively requires the sibling PR to land first OR sticks until next regeneration. In practice the regenerated file in this PR includes the constant, so the build works today; the sibling PR keeps it working tomorrow.

## Related

- Closes #952
- #932 (blob-path equivalence verb — closed; foundation this builds on)
- #957 (\`EVICTION_MODELS=ON\` build break — closed; the manual workaround that gets removed by the re-import)
- #961 (XGBoost latency target — separate; affects whether to graduate at all)
- #953 (xgb-as-default decision gate — downstream of this PR + #961)


johnjezl:
## Pi 5 hardware verification (pi-5-2, Cortex-A76 @ 2.4 GHz, 2026-05-17)

Phase 3 \`--baked\` mode verified on real silicon:

\`\`\`
slmos> bench xgb-equiv-evict --baked 0:/slmstore/test_vectors_xgb_evict.bin 0:/slmstore/expected_evict.bin
xgb-equiv-evict[baked]: 1000/1000 match (tol=1e-4)   2048 ns/predict   PASS
\`\`\`

**1000/1000 match, 2,048 ns/predict.**

Worth a callout: the baked-path latency is ~**687× faster** than the blob path (the prior #932 verification logged 1,407,823 ns/predict for the same 1000-vector corpus). The reason is the FFI overhead — the blob FFI does \`active_blob() + parse_payload()\` on every call, while the baked FFI is a direct function call into the compiled if-else chain.

That's good context for #961 (the < 1 µs hot-path target ticket): in the *actual eviction hot path* (\`select_victim\` → \`xgboost_score\` → eventually \`generated::xgb_predict\` or a cached parsed model), latency is closer to the baked number than the blob number. The 4,117 ns / 6,098 ns numbers from #962's per-policy latency table are for full \`select_victim\` cycles with feature extraction and policy bookkeeping, not just the predict call.

Test plan checkbox: Pi 5 hardware ✅. All four checks now pass; PR ready for merge.

Issue #1002, opened 2026-05-25
hailo: BAR4 fw-memory inspection — find fw-internal state divergence post-#1001


johnjezl:
## Context

Follow-up to #1001 (host-side trace-diff complete, wedge proven fw-internal).

The post-PR-#999 trace diff against HailoRT MNIST shows the first 9 BAR2 writes byte-identical between SLM-OS and HailoRT. After write 9, SLM-OS polls ch=2 NUM_PROC 2369 times reading `0x00000000` while HailoRT continues posting transfers because fw consumes on its side. **No host-side op HailoRT does that SLM-OS doesn't** — the wedge is fw-internal state, not host MMIO.

The next host-comparable move is reading fw's internal state via BAR4 to find what differs between fw's healthy (HailoRT-consuming) state and fw's wedged (SLM-OS-stuck) state.

## Tooling that already exists

`kernel/inference/inference_device_hailo.c:2434` already has `hailo_bar4_snap(hailo_bar4_snap_pre)` wired in under `HAILO_WIRE_DEBUG`. The infrastructure is there; what's missing is:

1. A matching `hailo_bar4_snap_post` at the post-timeout point in `hailo_backend_run` (the function already does plenty of post-fail diagnostics at lines 2466-2492 — adding a BAR4 snap is one more line).
2. A diff harness that compares pre-submit vs post-timeout BAR4 content and prints addresses + before/after values.
3. A way to take the matching snapshot on the HailoRT-on-Linux side. The instrumented `hailo_pci` (PR #997) already traces BAR4 reads/writes; we'd want to instrument it to also dump the full BAR4 region at the same lifecycle points.

## What to look for

Inside fw, every active boundary channel has a tracking structure: per-channel num_avail, num_proc, descriptor list IOVA, credit count, data_id, stream_index, batch state, scheduler-task-queue entry. fw writes some of these to BAR4-mapped memory regions HailoRT (and the kernel driver) can read.

Candidates to scan for in the diff:

- **Channel-state table.** If ch=2's slot shows `armed=false` or `credits=0` post-load, fw didn't process OPEN_BOUNDARY_INPUT_CHANNEL correctly even though the RPC returned `rc=0`. This would be the smoking gun for the LCU-template-might-be-incomplete hypothesis at `hailo_cs_translator.c:354-360`.
[...]

Issue #1003, opened 2026-05-25
hailo: bisect SAGE1_MIPI_RX_13 ECC trigger — byte-diff SLM-OS vs HailoRT CLEAR_APPS RPC


johnjezl:
## Context

Follow-up to #1002 (ECC ↔ wedge inseparable across N=9). The 10-boot quantification showed:

- 9/9 valid boots wedge
- 4-6 CPU_ECC_ERROR + 5-6 CPU_ECC_FATAL events per boot, all on `memory_bitmap=0x00001000 = SAGE1_MIPI_RX_13`
- Linux HailoRT on the same Pi 5 + same MNIST HEF: **zero CPU_ECC events**

The per-RPC D2H notification drain captured in the first BAR4 inspection run (#1002 comment 4537695227) shows the trigger sequence:

```
post RESET:                              0 events  ← baseline, no ECC
post CLEAR_CONFIGURED_APPS:              1 event   ← FIRST ECC fires here
post GET_HW_CONSTS:                      1 event
post SET_NETWORK_GROUP_HEADER:           1 event
post SET_CONTEXT_INFO (each of 4):       1 event
post CHANGE_STATUS(ENABLED):             1 event
post-ENABLED settle pings:               1 event
```

**`CLEAR_CONFIGURED_APPS` (opcode 0x47, sent to CORE CPU) is the first ECC trigger.** Subsequent ECCs may be cascading effects. Linux calls the same RPC at the same lifecycle point without triggering any ECC.

## Hypothesis

SLM-OS sends CLEAR_APPS in a slightly-different wire form (md5 prefix, buffer_len, or buffer body bytes) than HailoRT, leaving fw with a stale or wrong-format state expectation. When fw subsequently touches the SAGE1_MIPI_RX_13 memory region, it reads garbage and the ECC checker reports the corruption. The actual bit-flip didn't happen — SLM-OS just made fw read from somewhere it shouldn't have, which the ECC engine treats as a faulted access.

If true, finding the byte-level RPC divergence and fixing it should eliminate the ECC trigger, which should eliminate the wedge (per #1002's tight correlation).

## What's already known

- SLM-OS CLEAR_APPS RPC builder lives in `kernel/ai_accel/hailo/hailo_control.c`. Search for `HAILO_CONTROL_OPCODE_CONTEXT_SWITCH_CLEAR_CONFIGURED_APPS` (= 0x47).
[...]

PR #999, opened 2026-05-25
hailo: fix VDMA boundary channel direction mismatch wedging IN-channel inference (#682)

johnjezl:
## Summary

- Two host-MMIO call sites (`hailo_shell.c`, `inference_device_hailo.c`) hardcoded `.input_channel = 0, .output_channel = 1`, while the CS translator emits `packed_vdma_channel_id = config + {1, 15}` (ch=2 H2D, ch=16 D2H) in the `OPEN_BOUNDARY_{INPUT,OUTPUT}_CHANNEL` RPCs the firmware consumes. The host and the firmware were referencing different MMIO registers on different channels and never converging — the IN-channel wedge that's been #682's signature.
- Derive both channel indices at both call sites from `HAILO_CS_DEFAULT_CONFIG_VDMA_CHANNEL + HAILO_CS_BOUNDARY_{INPUT,OUTPUT}_CHANNEL_OFFSET` so the host-side MMIO and the CS RPC share a single source of truth.
- Add a runtime H2D [0, 15] / D2H [16, 31] direction-range check in `hailo_infer_run` so future direction-mismatched callers fail fast with `HAILO_ERR_INVAL` instead of silently wedging.

## Root cause

Located 2026-05-25 via the first side-by-side BAR2 diff between HailoRT-on-Linux and SLM-OS-native running the same MNIST HEF on pi-5-1:

|  | HailoRT (Linux) | SLM-OS (before fix) |
|---|---|---|
| BAR2 ops | 98 | 4800 |
| Channels touched | **ch=2 (H2D) + ch=16 (D2H)** | **ch=0 + ch=1** |
| `NUM_PROC` behaviour | non-zero progressing | 0x00000000 forever |

The translator at `hailo_cs_translator.c:212/249` already computes the right packed channels from `config_vdma_channel + offset` constants. The header pins the H2D/D2H direction split with `_Static_assert`s. The bug was purely that the host-side VDMA-arming code at the two call sites bypassed those constants and used raw \"first available\" indices.

## Regression coverage

`kernel/tests/test_hailo.c`:

- `test_infer_rejects_input_in_d2h_range` — input in [16, 31] is rejected
- `test_infer_rejects_output_in_h2d_range` — output in [0, 15] is rejected
- `test_infer_rejects_old_default_pair` — pins the shipped `(0, 1)` value to fail
- `test_infer_translator_offsets_yield_valid_channels` — pins the derivation chain so a future bump to either constant that pushes a boundary into the wrong direction trips the test before it can ship

[...]

PR #1004, opened 2026-05-26
hailo: match HailoRT load-time control-RPC cadence (4 unjustified divergences)

johnjezl:
## Summary

SLM-OS had four small divergences from HailoRT's load-time control-RPC cadence — none of them load-bearing, none justified by a documented rationale, all visible in the HailoRT wire captures we already have on disk. Each was a risk surface: \"fw responds differently to SLM-OS than to HailoRT\" was a permanent variable confounding the #1002 / #1003 investigation. This PR removes the four divergences as a hygiene change.

**Not a fix for #1002.** The remaining wedge survives all four changes individually (each tested in #1003). This PR is about making future ECC investigation cleaner by removing easy-to-dismiss confounders, not about solving the wedge.

## The four changes

Each is one commit, each cites the corresponding line in the checked-in HailoRT wire capture.

### 1. \`control_next_sequence\` returns 0

HailoRT sends \`sequence=0\` on every control RPC across 27 captures in two wire dumps (\`hailort-v4.23.0-wire-capture-{mnist-pi5,mobilenet}.txt\`). SLM-OS atomically incremented. SLM-OS doesn't validate response sequence, so the increment had no consumer inside SLM-OS — purely a wire divergence.

### 2. Drop pings between RESET and CLEAR_APPS

HailoRT issues CLEAR_APPS 127 µs after CHANGE_STATUS(RESET) with no RPC between (mobilenet capture, t=393.529554 → 393.529681). SLM-OS inserted \`context_switch_settle_pings(\"CLEAR_APPS\")\` + \`hailo_core_cpu_settle()\`. The surrounding comment claimed to mirror HailoRT's cadence; the actual wire capture proves it doesn't here.

### 3. Two GDIs before RESET (was one)

HailoRT issues GET_DEVICE_INFO twice before CHANGE_STATUS(RESET) (mobilenet capture: GDI at 393.529161 + 393.529358, then RESET at 393.529554). SLM-OS's shared \`context_switch_settle_pings\` helper emits one GDI per call site. Added an inline second GDI specifically at the pre-RESET site so the cadence matches; other call sites left alone (their HailoRT-matching cadence varies and isn't addressed here).

### 4. GET_HW_CONSTS 4× not 6×

HailoRT calls GET_HW_CONSTS exactly four times back-to-back during init (verified in both wire captures: 4 occurrences of opcode \`00 00 00 48\`). The previous SLM-OS comment claimed \"SIX times\" citing a \"Pi OS inference trace 2026-05-09\" that isn't reproducible from the checked-in captures.

## Tests

- \`make test\` — full QEMU suite passes (one test pinned the old \`14u\` core-RPC count; updated to \`12u\` to match the new GET_HW_CONSTS=4 count).
- Hardware-verified on pi-5-1 (Pi 5 + AI HAT+):
  - Load completes (\`model loaded (handle=1), ai_policy_hailo armed\`)
  - GET_HW_CONSTS prints 4 iterations as expected
  - ECC count within noise of baseline (4 ECC_err + 5 ECC_fatal vs #1002 N=9 baseline avg 4.4/5.2)
  - Wedge still reproduces (expected — #1003 disconfirmed each of these individually)

## Test plan

- [x] \`make test\` (QEMU ARM64)
- [x] Hardware build + boot + load + runmodel on pi-5-1
- [ ] Reviewer spot-check that the wire-capture citations match what's actually in the referenced files
- [ ] Reviewer judgment on whether the inline second GDI (commit 3) should instead modify the shared \`context_switch_settle_pings\` helper to take a count parameter

## References

- #1002 — ECC ↔ wedge inseparable across N=9 boots
- #1003 — host-side hypotheses disconfirmed; this PR is the cleanup from that investigation
- \`~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mnist-pi5.txt\`
- \`~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mobilenet.txt\`

### DIFF
--- a/kernel/ai_accel/hailo/hailo_control.c
+++ b/kernel/ai_accel/hailo/hailo_control.c
@@ -96,7 +96,19 @@ static uint32_t control_sequence = 0;
 
 static inline uint32_t control_next_sequence(void)
 {
-    return __atomic_fetch_add(&control_sequence, 1, __ATOMIC_RELAXED);
+    /* HailoRT v4.23 sends sequence=0 on every control RPC. Verified
+     * across 27 captures in two wire dumps:
+     *   ~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mnist-pi5.txt
+     *   ~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mobilenet.txt
+     * — every `hailo-req` shows bytes 8..11 of the common_header as
+     * `00 00 00 00`. SLM-OS doesn't validate response sequence (the
+     * `control_check_response_header` path matches by opcode and
+     * response_len, not sequence), so incrementing has no consumer
+     * inside SLM-OS and is a pure wire divergence from the reference.
+     * Match the reference. The static counter is kept (zeroed) for
+     * potential future debug correlation if needed. */
+    (void)control_sequence;
+    return 0;
 }
 
 /*
--- a/kernel/ai_accel/hailo/hailo_control.c
+++ b/kernel/ai_accel/hailo/hailo_control.c
@@ -84,30 +84,15 @@ static inline uint32_t hailo_be32_to_cpu(uint32_t v)
     return __builtin_bswap32(v);
 }
 
-/* Sequence counter. Firmware echoes this back in the response header
- * so a response can be correlated with a request; we check it against
- * what we sent. Incremented atomically per send so hailo_control_*
- * callers that race in the request-build phase (before taking
- * control_lock inside send_recv) can't produce colliding sequence
- * numbers — otherwise two CPUs could read the same pre-increment
- * value, send distinct requests with identical seq, and each mistake
- * the other's echoed-back response for their own. */
-static uint32_t control_sequence = 0;
-
 static inline uint32_t control_next_sequence(void)
 {
     /* HailoRT v4.23 sends sequence=0 on every control RPC. Verified
-     * across 27 captures in two wire dumps:
-     *   ~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mnist-pi5.txt
-     *   ~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-mobilenet.txt
-     * — every `hailo-req` shows bytes 8..11 of the common_header as
-     * `00 00 00 00`. SLM-OS doesn't validate response sequence (the
-     * `control_check_response_header` path matches by opcode and
-     * response_len, not sequence), so incrementing has no consumer
-     * inside SLM-OS and is a pure wire divergence from the reference.
-     * Match the reference. The static counter is kept (zeroed) for
-     * potential future debug correlation if needed. */
-    (void)control_sequence;
+     * across 27 captures in two wire dumps
+     * (~/slmos-ref/derivatives/hailort-traces/hailort-v4.23.0-wire-capture-{mnist-pi5,mobilenet}.txt):
+     * every `hailo-req` shows bytes 8..11 of the common_header as
+     * `00 00 00 00`. SLM-OS doesn't validate response sequence —
+     * `control_check_response_header` matches by opcode + response_len.
+     * Return 0 to match the reference. */
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


johnjezl:
## Update — added 3 more matching commits (PR now 6 total)

Following the systematic byte-diff of all 7 load-time RPCs against HailoRT's MNIST wire capture, the remaining divergences in SET_NETWORK_GROUP_HEADER's application_header were addressed.

### Byte-diff results

| RPC | Status |
|---|---|
| CHANGE_STATUS(RESET) | ✓ byte-identical to HailoRT |
| CLEAR_APPS | ✓ byte-identical |
| GET_HW_CONSTS × 4 | ✓ byte-identical (count fixed by commit 4) |
| **SET_NETWORK_GROUP_HEADER** | **had 3 divergences in application_header** |
| SET_CONTEXT_INFO(ACTIVATION) | ✓ identical modulo 4 IOVA bytes |
| SET_CONTEXT_INFO(BATCH_SWITCHING) | ✓ **byte-identical** (114/114 bytes) |
| SET_CONTEXT_INFO(PRELIMINARY) | ✓ identical modulo 8 IOVA bytes |
| SET_CONTEXT_INFO(DYNAMIC) | ✓ identical modulo 4 IOVA bytes |
| CHANGE_STATUS(ENABLED) | ✓ byte-identical |

The 16 total byte divergences across SET_CONTEXT_INFO bodies all decode to `dma_address` LSBs in `host_buffer_info` blocks — SLM-OS PMM picks lower IOVAs (~0x1002_xxxx_xxxx) than Linux's DMA allocator (~0x103a_xxxx_xxxx), but the same window prefix. Not a bug.

### New commits 5 + 6

5. **`preliminary_run_asap=1` + `can_fast_batch_switch=1`**: HailoRT MNIST wire capture shows both = 1. SLM-OS sent 0/0 based on a 2026-04-23 test that "did not fix the boundary-submit silence". That test ran with the channel-index bug (#682, fixed by PR #999 2026-05-25) still in place — the wedge was going to fire regardless of these flags. The test's conclusion ("ASAP mode needs paired ActivateChannel actions SLM-OS doesn't emit") is contradicted by the PRELIMINARY byte-diff: SLM-OS's PRELIMINARY action stream is byte-identical to HailoRT's, and HailoRT runs with asap=1 + that exact stream.

6. **`external_action_list_address = 0` (was `0xFFFFFFFF`)**: HailoRT MNIST wire capture shows 0. SLM-OS's prior 0xFFFFFFFF was based on a code comment claiming "0 would be rejected by firmware" — the wire capture directly contradicts that. The `HAILO_CS_NO_DDR_ACTION_LIST` sentinel constant is left in the header for future paths that might genuinely need it.

### Hardware verification (PR #1004 v2)

Ran the full 6-commit branch on pi-5-1 (Pi 5 + AI HAT+):

```
LOAD: ECC_err=4 ECC_fatal=5   (baseline 4.4/5.2 — within noise)
RUN:  ECC_err=0 ECC_fatal=0
GET_HW_CONSTS calls: 4         (matches HailoRT, per commit 4)
Model loaded: OK
Wedge:        still reproduces (expected — matching, not fixing)
```

Load completes cleanly. ECC count unchanged from baseline. Wedge still fires (as predicted — every individual change in this PR was disconfirmed as a #1002 fix in #1003).

### Net result

**SLM-OS's load-time control RPC stream is now byte-faithful to HailoRT v4.23 for MNIST, modulo expected IOVA variation.** The #1002 wedge investigation can no longer be confounded by load-time wire-byte divergences from the reference — that whole class of "easy" candidates is closed off.

Test plan box ticked: `make test` passes (one test pinned the previous `HAILO_CS_NO_DDR_ACTION_LIST` expectation; updated to `0`).

johnjezl:
## Update — commit 7 added: inter-RPC settle floor 50ms → 250us

Investigating the wedge against the matched baseline surfaced one more unjustified divergence with a now-disproven rationale.

### What was there

`HAILO_CORE_CPU_SETTLE_FLOOR_US = 50000` (50ms) enforced a 50ms minimum gap between every CORE-CPU RPC during load. The 2026-05-09 commit 8b2145ad that introduced this justified it with:

> "Linux's HailoRT achieves the gap incidentally via user-space scheduling latency (seconds between RPCs from vstreams library)"

### What the wire capture shows

HailoRT v4.23 MNIST wire capture timestamps:

```
393.528893  IDENTIFY
393.529161  GDI            (+268 us)
393.529358  GDI            (+197 us)
393.529554  CHANGE_STATUS  (+196 us)
393.529681  CLEAR_APPS     (+127 us)
```

**~100-300 us between RPCs, not seconds.** The original premise is wrong. SLM-OS was 250× slower than the reference. The 50 vs 200 ms bisect that justified keeping 50ms never tested below 50ms — it just compared two values that were both already 100× slower than Linux.

### Verification on pi-5-1

```
LOAD: ECC_err=4 ECC_fatal=5   (baseline 4/5)
Wedge: still reproduces       (expected)
Load latency: ~300 ms faster  (saved 49.75 ms × 6 settle sites)
```

ECC and wedge **unchanged** — consistent with the original commit's finding that "inter-RPC gap is not a clean lever on the boundary-channel wedge." But the previous floor was a 250× divergence from the reference for no actual benefit, so this commit brings the inter-RPC pacing in line.

### PR #1004 final state (7 commits)

| # | Commit | Effect |
|---|---|---|
| 1 | sequence=0 | matches wire format |
| 2 | drop pings between RESET and CLEAR_APPS | matches cadence |
| 3 | add 2nd GDI before RESET | matches cadence |
| 4 | GET_HW_CONSTS 4× not 6× | matches count |
| 5 | preliminary_run_asap=1 + can_fast_batch_switch=1 | matches flags |
| 6 | external_action_list_address=0 | matches field value |
| 7 | inter-RPC floor 50ms → 250us | matches timing |

SLM-OS's load-time control path is now byte-faithful AND timing-faithful to HailoRT within the limits of the wire-capture data we have. None of the 7 individual changes fix #1002 (each was tested in #1003 or earlier and disconfirmed), but together they close out the entire class of "load wire/cadence divergence" as a confounder for future ECC / boundary-channel investigation.

`make test` passes. Hardware verified on pi-5-1.

Issue #1005, opened 2026-05-27
hailo: byte-diff CCW data content SLM-OS vs HailoRT (post-#1004 next wedge candidate)


johnjezl:
## Context

Follow-up to #1003 (ECC inseparable from wedge across N=9) and PR #1004 (load-time control RPC stream byte-faithful to HailoRT modulo IOVA). With every load-time wire and cadence divergence now closed, the wedge still reproduces with the same SAGE1_MIPI_RX_13 ECC pattern.

The next plausible angle that hasn't been directly compared: **the CCW data content fw reads via DMA**.

## What CCW is

The HEF carries a stream of \"context-switch workload\" actions that configure fw's NN engines for a network group. During load, the host puts these bytes into a DMA buffer and tells fw to pull them via the **config VDMA channels** (ch=0 and ch=1 on Hailo-8 PCIe). The fw reads each CCW action and uses it to program internal NN cluster state (LCUs, sequencers, MIPI input config, etc.).

SLM-OS reports the CCW breakdown at load time:

\`\`\`
ccw: 28 action(s), total 112256 bytes
  cfg_channel[0]: 22 action(s), 55792 bytes
  cfg_channel[1]: 6 action(s), 336 bytes
\`\`\`

These bytes are written to host DMA buffers, then pulled by fw via the config channels (which DO advance num_proc as fw consumes — `[vdma] ch=0 wait_proc done after 200 us proc=0x006e006d`). So the CCW upload mechanically works. What we don't know is whether the CCW *content* SLM-OS hands fw matches what HailoRT hands fw.

## Why this matters

CCW is the only path through which fw learns about NN-cluster-specific config (including MIPI input routing — which is what SAGE1_MIPI_RX_13 is). If SLM-OS's HEF parser extracts CCW bytes differently from HailoRT's (different field interpretation, byte order, sub-action layout), fw configures its MIPI receivers with garbage values. When fw later validates those values during CLEAR_APPS / other RPCs, it reads uninitialized SAGE1_MIPI_RX_13 → ECC fires.

This hypothesis fits the observed ECC pattern: events fire on RPCs that touch NN-cluster state (CLEAR_APPS validates the current configuration, SET_CONTEXT_INFO reads it back). RESET doesn't trigger because RESET *zeroes* fw state without reading it.

## What we have

[...]

PR #1007, opened 2026-05-27
hailo: CCW fingerprint dump under HAILO_WIRE_DEBUG (#1005 diagnostic)

johnjezl:
## Summary

Adds a diagnostic dump of the assembled CCW upload buffers at HEF load time, gated under \`HAILO_WIRE_DEBUG\`. The fingerprint per channel:

- First 64 + last 64 bytes of the buffer (hex)
- Per-action summary: \`cfg_channel_index\`, proto-action-index, CCWS offset, byte size

For MNIST this produces ~30 lines of output per load — small enough not to drown the boot log, large enough to spot-check that SLM-OS's HEF parser produces the expected \`(offset, size, channel)\` tuples and that the slice extraction (\`copy_ccws_for_cfg_channel\`) emits the bytes the HEF actually contains.

Used in #1005 iteration 1 (comment 4550551964) to confirm SLM-OS's CCW upload is byte-faithful to the raw HEF at the parsed offsets:

```
Ch=1 action 0 (CCWS off=111920 size=16):
  SLM-OS head:  00 00 00 00 84 06 20 00 00 00 00 02 00 00 00 00
  HEF @ 240533: 00 00 00 00 84 06 20 00 00 00 00 02 00 00 00 00  ✓
```

## Why a separate PR

The remaining #1005 work (comparing SLM-OS's action list against HailoRT's for the same HEF) requires a Linux-side capture and isn't addressed here. The SLM-OS-side fingerprint this commit produces is the artifact the eventual cross-side diff will compare against — landing it in a dedicated commit keeps the diagnostic available without conflating with the analysis work.

## Test plan

- [x] \`make test\` passes (no behavior change in default builds)
- [x] Hardware build + load on pi-5-1 with \`HAILO_WIRE_DEBUG=ON\` — fingerprint output verified against raw HEF file
- [x] Dump artifact saved to \`~/slmos-ref/derivatives/slmos-traces/slmos-ccw-dump-mnist-pi-5-1-2026-05-26.log\`
- [x] No runtime cost in default builds (entire dump block is inside \`#ifdef HAILO_WIRE_DEBUG\`)

## References

- #1005 — parent issue
- PR #1004 — load-time wire/cadence cleanup that surfaced CCW content as the next angle
- #1002, #1003, PR #999 — full investigation chain

### DIFF
--- a/kernel/inference/inference_device_hailo.c
+++ b/kernel/inference/inference_device_hailo.c
@@ -956,6 +956,68 @@ static int copy_ccws_for_cfg_channel(
 }
 
 #ifdef HAILO_WIRE_DEBUG
+/* #1005: assembled-CCW fingerprint for offline diff vs HailoRT.
+ *
+ * Called by `context_switch_load` immediately after each per-channel
+ * CCW buffer is filled by `copy_ccws_for_cfg_channel`. Prints:
+ *   - First `HAILO_CCW_DUMP_FINGERPRINT_BYTES` bytes of the buffer (hex)
+ *   - Last `HAILO_CCW_DUMP_FINGERPRINT_BYTES` bytes if the buffer is
+ *     larger than 2× the fingerprint size (avoids head/tail overlap)
+ *   - Per-action summary (channel, proto-action-index, CCWS offset,
+ *     byte size) filtered to `channel_filter`.
+ *
+ * `channel_filter` semantics:
+ *   - Pass HAILO_CCW_DUMP_CHANNEL_ANY to list every action regardless
+ *     of cfg_channel_index. Used for the single-channel primary buffer
+ *     (a verbatim copy of the whole CCWS block — all actions apply).
+ *   - Pass a concrete channel index (typically 0 or 1) to list only
+ *     actions tagged for that channel. Used for both buffers in the
+ *     dual-channel path: primary carries cfg_channel_index=1, secondary
+ *     carries cfg_channel_index=0. */
+#define HAILO_CCW_DUMP_FINGERPRINT_BYTES 64u
+#define HAILO_CCW_DUMP_CHANNEL_ANY       UINT32_MAX
+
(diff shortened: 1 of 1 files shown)


johnjezl:
## Update — addressed all 3 review items

**Warning (unbraced `for` body)**: Fixed. Both `for (...) uart_printf(...);` single-line forms are now multi-line with braces. Same fix in the helper (where the loops now live).

**Suggestion 1 (factor helper)**: Done. Extracted `hailo_ccw_dump_fingerprint(label, hef, channel_filter, buf, bytes)` near `copy_ccws_for_cfg_channel`. Both dump call sites now collapse to a single function call. Net diff is +75 lines (was +60) because the helper carries its own ~50 lines, but the two call sites went from ~30 lines each to 3 lines each — easier to read inline, single place to maintain.

**Suggestion 2 (named constant)**: Done. `HAILO_CCW_DUMP_FINGERPRINT_BYTES = 64u` at file scope; used 4 places (head clamp, head label, tail threshold `2u * `, tail label).

**Bonus**: While factoring, also added `HAILO_CCW_DUMP_CHANNEL_ANY = UINT32_MAX` sentinel so the single-channel-mode primary call (where the buffer holds a verbatim copy of the whole CCWS block) can list every action regardless of `cfg_channel_index`. Original inline code preserved this behavior via `if (use_dual && ...)`; helper preserves it explicitly via the sentinel.

### Verification

- `make test` passes
- Pi 5 build with `HAILO_WIRE_DEBUG=ON` compiles
- Output format unchanged from the original inline version (verified by mental diff against the captured `~/slmos-ref/derivatives/slmos-traces/slmos-ccw-dump-mnist-pi-5-1-2026-05-26.log`)
- No runtime cost in default builds (helper, both call sites, and the constants are all inside `#ifdef HAILO_WIRE_DEBUG`)

Force-pushed as `92048f36`.

